PR #2, opened 2015-05-07
Support g (jump to the top) and G (jump to the bottom)

frantic:
Super handy, `less` and `vi` support it

### DIFF
--- a/src/screenControl.py
+++ b/src/screenControl.py
@@ -316,10 +316,13 @@ class Controller(object):
         self.moveIndex(-pageHeight)
 
     def moveIndex(self, delta):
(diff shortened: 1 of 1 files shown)


pcottle:
Awesome! Do you think its worth adding to the usage bar as well? Bit worried about space....

frantic:
Was thinking about that as well. For now I think it's okay to leave it undocumented, people who got used to `less`/`vi` style g/G will be pleasantly surprised :)

pcottle:
> For now I think it's okay to leave it undocumented, people who got used to less/vi style g/G will be pleasantly surprised :)

Yeah I agree -- hopefully we don't go too far down the `vim` emulation rabbit hole

Issue #4, opened 2015-05-07
Python 3.0 compatibility 


dannguyen:
I'm on 10.10.3. Ran `brew install fpp` and then `fpp`, and got this:

```
File "/usr/local/Cellar/fpp/0.5.3/libexec/src/processInput.py", line 126
    print USAGE_STR
                  ^
SyntaxError: Missing parentheses in call to 'print'
  File "/usr/local/Cellar/fpp/0.5.3/libexec/src/choose.py", line 54
    print 'Nothing to do!'
                         ^
SyntaxError: Missing parentheses in call to 'print'
sh: /Users/dan/.fbPager.sh: No such file or directory
```

pcottle:
What version of python are you running? `python --version`. It's likely a Python 3.0 compatibility issue

dannguyen:
Ah yes, I'm on Python 3.x

pcottle:
PR #15 helped this a bit as well in the meantime -- @dannguyen @vimeitor have either of you tried that yet?

Let me see what the python 2to3 converter says...

vimeitor:
@pcottle I can confirm that it works very well. It's a good first step for sure :).

sdornan:
Would love for this to be compatible with Python 3.

PR #5, opened 2015-05-07
Use python2 instead of python

vimeitor:
A lot of Linux distributions use python3 instead of python2 as the
default version of python. Let's be more specific.


pcottle:
Hrm, this wouldn't work for a lot of mac setups that don't have the python2 command (since python 2.7 is the default) so this can't go in as-is

Do we want to instead handle this in Issue #4? If we support both 3 and 2, we could get the best of both worlds.

vimeitor:
Oh, sorry about that. Yes, being able to run it with both versions of python would definitely be great.

Issue #6, opened 2015-05-07
Git installation troubles


cboden:
I have a couple issues with the installation via git. I'm on OS X 10.10.3 with Python 2.7.6.

1) The installation instructions create a cyclic symbolic link: `ls -ltr | tail -n 1`
> lrwxr-xr-x (etc) fpp -> ./fpp

Adjusting the `ln` command to use the full path for the source fixes that.

2) After the symbolic link is setup correctly I get the following error when running `fpp --help`:
> /usr/local/bin/fpp: line 18: cd: /usr/local/bin//usr/local/PathPicker: No such file or directory
python: can't open file '/src/printHelp.py': [Errno 2] No such file or directory

I have placed PathPicker in `/usr/local/PathPicker` and linked the binary to `/usr/local/bin/fpp`.

kkoch986:
had the exact same experience, fixed by changing 

```
while [ -h "$WHEREAMI" ]; do
  WHEREAMI=$(dirname "$WHEREAMI")"/"$(readlink "$WHEREAMI")
done
```

to 

```
while [ -h "$WHEREAMI" ]; do
  WHEREAMI=$(readlink "$WHEREAMI")
done
```

frantic:
@cboden, @kkoch986 - could you please try #16 and see if it works for you?

kkoch986:
works for me

cboden:
works for me as well

PR #10, opened 2015-05-07
Fix $whereami and other bash linting for Issue #6

n06:
Fixes an issue where the path of the shell script is added on twice. (Eg. <path>/<path>/fpp). Also fixed some variable scoping in the script just for niceness.

### DIFF
--- a/fpp
+++ b/fpp
@@ -12,7 +12,7 @@ WHEREAMI=$0
 # Follow until we get our symlink resolved, since
 # homebrew has multiple hops.
 while [ -h "$WHEREAMI" ]; do
-  WHEREAMI=$(readlink "$WHEREAMI")
+  WHEREAMI=$(dirname "$WHEREAMI")"/"$(readlink "$WHEREAMI")
 done
 BASEDIR=$(dirname "$WHEREAMI")
 BASEDIR=$(cd "$BASEDIR" && pwd)
--- a/fpp
+++ b/fpp
@@ -12,10 +12,10 @@ WHEREAMI=$0
 # Follow until we get our symlink resolved, since
 # homebrew has multiple hops.
 while [ -h "$WHEREAMI" ]; do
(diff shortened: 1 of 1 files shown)


pcottle:
definitely down for this line:
`+BASEDIR=$(cd "$BASEDIR" && pwd)`

however i remember having to append the slashes since the `brew` installation has multiple hops. Let me get some command line output

pcottle:
```
[pcottle:~/Dropbox (Facebook)/wip/PathPicker:master]$ git diff HEAD~10 --stat | fpp
WHEREAMI is /usr/local/bin/../Cellar/fpp/0.5.3/bin/fpp
WHEREAMI is /usr/local/bin/../Cellar/fpp/0.5.3/bin/../libexec/fpp
nothing to do!
```

Yeah, see the tricky thing is that `homebrew` installs **relative** symbolic links, so switching to your first change unfortunately breaks:

```
[pcottle:~/Dropbox (Facebook)/wip/PathPicker:master]$ git diff HEAD~10 --stat | fpp
WHEREAMI is ../Cellar/fpp/0.5.3/bin/fpp
/usr/local/bin/fpp: line 19: cd: ../Cellar/fpp/0.5.3/bin: No such file or directory
python: can't open file '/src/processInput.py': [Errno 2] No such file or directory
python: can't open file '/src/choose.py': [Errno 2] No such file or directory
nothing to do!
```

pcottle:
Do you mind updating with just the `BASEDIR=$(cd "$BASEDIR" && pwd)` change? ill happily accept that! that was just my shortsightedness / bash inexperience

n06:
Ah I see, the perils of being on linux and not OS X. I'll change this to just include the scoping for now, and see if I can come up with something else for the path issues later

pcottle:
@n06 I actually think this got solved in #16 since we have better symbolic link resolution everywhere. do you mind checking if that works for you? closing out in the meantime since this is on a stale rev

PR #13, opened 2015-05-07
fix horizontal spacing for index.html

SteveByerly:
The current index.html has a 15px white margin off the right side of the page.

### DIFF
--- a/assets/launch_page.css
+++ b/assets/launch_page.css
@@ -14,7 +14,7 @@
   padding-bottom: 16px;
 }
 
-.jumbotron .container div {
+.jumbotron .heading {
   text-align: center;
   padding: 40px 16px;
 }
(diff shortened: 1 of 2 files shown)


pcottle:
Thanks for reducing the CSS specificity -- thats a habit I have from Facebook where our CSS can be downloaded in any order (so same-specificity rules are a recipe for bugs). Ours are deterministic though

Can you point out where you are seeing the extra spacing? Or what browser you are using? I don't see any differences on chrome...

screen shot 2015-05-07 at 12 02 36 pm
screen shot 2015-05-07 at 12 02 55 pm

SteveByerly:
The issue is that rows need to be nested somewhere in a container. The row in the jumbotron was outside of the container. Based on how the html and css were structured, i updated as made sense to me.

screen shot 2015-05-07 at 12 07 06 pm

pcottle:
Ah I see -- this is a bit more semantically correct so sure I'll pull! Let me resolve the merge conflict I just introduced

PR #14, opened 2015-05-07
Use default terminal colors

frantic:
RFC. When using default colors, fpp output matches terminal's color scheme and looks better. See [`curses.use_default_color`](https://docs.python.org/2/library/curses.html#curses.use_default_colors)

Using solarized-light color scheme:

before:

screen shot 2015-05-07 at 11 46 21 am

after:

screen shot 2015-05-07 at 11 46 28 am

I've also tried it in localhost tmux session, seems to be running without issues.

### DIFF
--- a/src/screenControl.py
+++ b/src/screenControl.py
@@ -214,6 +214,7 @@ class Controller(object):
         self.numMatches = len(self.lineMatches)
 
         self.setHover(self.hoverIndex, True)
+        curses.use_default_colors()
         logger.addEvent('init')
 
     def getScrollOffset(self):


pcottle:
Do you think the default colors of red / blue will look OK when using whatever terminal colors we have?

I agree this is a big step forward though when using the default colors though, otherwise its **super** visually jarring. Glad we already hired you @frantic haha

:tada: :fireworks:

frantic:
> Do you think the default selection colors of red / blue will look OK when using whatever terminal colors we have?

Many other apps use red and blue colors, so if background doesn't play nicely with these - other apps will look bad as well (git, hg, anything else that does colors).

PR #16, opened 2015-05-07
Fix symlink resolution

### LINKED ISSUE #6: Git installation troubles

frantic:
I have fpp repo checked out locally, the binary is symlinked to my
`~/bin` which is also a symlink to some other place. Running `fpp`
used to show "No such file or directory".

I took this code from the internet and it seems to do the job well.

Fixes #6

### DIFF
--- a/fpp
+++ b/fpp
@@ -7,15 +7,14 @@
 # of patent rights can be found in the PATENTS file in the same directory.
 #
 # get the directory of this script so we can execute the related python
-WHEREAMI=$0
-
-# Follow until we get our symlink resolved, since
-# homebrew has multiple hops.
-while [ -h "$WHEREAMI" ]; do
-  WHEREAMI=$(dirname "$WHEREAMI")"/"$(readlink "$WHEREAMI")
+# http://stackoverflow.com/a/246128/212110
+SOURCE="${BASH_SOURCE[0]}"
+while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
+  BASEDIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+  SOURCE="$(readlink "$SOURCE")"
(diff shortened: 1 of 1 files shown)


cboden:
:+1: LGTM

pcottle:
:see_no_evil: it amazes me how hard resolving symlinks is. I tried a few variants of this until I landed on something that would work for both relative symlinks, absolute symlinks, multiple hops, etc etc.

A little concerned about our zsh users, since it doesnt seem to be equivalent:
http://stackoverflow.com/questions/9901210/bash-source0-equivalent-in-zsh

While we look for a way to use either, let me pull this down and try

pcottle:
Yeah this seems to work for all my tests locally. this is what I could find on the difference between `BASH_SOURCE` and `$0`:

>*convention* is that argv[0] is the name of the script, but it’s not necessarily. you can easily change $0 of a script by calling it with execv or something differently. in fact, the way busybox works is it’s this massive single binary that chooses its behavior based on what argv[0] is.

from:
http://www.mbeckler.org/blog/?p=152

So Lets go with `$0` so we keep zsh compatibility

PR #22, opened 2015-05-07
Update README.md

aledista:


sairion:
Following current instruction and executing `$ fpp` gives error reads "Too many levels of symbolic links" ... I made it work with `ln -s $(pwd)/fpp /usr/local/bin/fpp` but this works also ;)

pcottle:
@sairion the symbolic link jazz might have been fixed by #16, can you try that? I'm not sure if this is required after pulling in that updated.

@aledista my main concern for this is that the installation instructions would break if you cloned the repo as a different name. I do obviously list the cd command right after the clone, but in general this seems unnecessary since the symlink resolution should be able to handle the source being right there.

Going to close this one out since I think #16 fixes this, but let me know if after that it's still an issue!

sairion:
@pcottle Newest master (at the moment I write this) with the instruction still gives me same error, 

```
➜  ~  fpp
zsh: too many levels of symbolic links: fpp
```

(Actually I uses zsh, I should have said above.)

PR #23, opened 2015-05-07
fix BASEDIR on Linux

ninegene:
Fix the following script/base directory resolution issue in Linux.

```bash
$ ln -s ~/Projects/ninegene/PathPicker/fpp ~/bin/fpp

$ git log --name-status | fpp
/home/aung/bin/fpp: line 18: cd: /home/aung/bin//home/aung/Projects/ninegene/PathPicker: No such file or directory
python2: can't open file '/src/processInput.py': [Errno 2] No such file or directory
python2: can't open file '/src/choose.py': [Errno 2] No such file or directory
nothing to do!
```

### DIFF
--- a/fpp
+++ b/fpp
@@ -17,6 +17,10 @@ done
 BASEDIR=$(dirname "$WHEREAMI")
 BASEDIR=$(cd $BASEDIR && pwd)
 
+if [ "$(uname -s)" == "Linux" ]; then
+  BASEDIR=$(cd "$(dirname "$(readlink -f "$0")")" && pwd)
+fi
+
 for opt in "$@"; do
   if [ "$opt" == "--debug" ]; then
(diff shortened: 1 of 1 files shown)


bmwang:
this would break osx because -f is a gnu readlink flag, and osx uses posix readlink

pcottle:
@ninegene is this because of the differences in `readlink`? i'd actually rather introspect into the `readlink -f` option if thats possible rather than look at the operating system name

Also we just pulled #16 from the excellent @frantic, do you mind seeing if that resolves your situation? that should fix all the symlink resolution woes (I think this is on a stale rev).

Thanks a ton for the PR though! Closing out unless #16 doesnt fix your situation, then we can re-open

ninegene:
#16 works great. Thanks.

PR #24, opened 2015-05-07
pep8ed

aledista:


pcottle:
Awesome! thanks a bunch -- I used Google's yapf:
https://github.com/google/yapf

to get the original bulk of formatting down, but I agree with these changes so might as well be pep8 compliant!

@aledista is there any way to easily test if all files are pep8 compliant? id be nice to make this part of the build step

pcottle:
I found https://github.com/jcrocholl/pep8/ hehe. I added a simple build script just so this doesnt regress (and neither do the tests) -- thanks for the help again!

aledista:
@pcottle, I'm glad you found my commit helpful! For style conventions I generally use https://flake8.readthedocs.org

PR #40, opened 2015-05-08
x mode

xavierbeynon:
x mode labels lines so that files can be selected without scrolling.  Enter x mode by pressing x, select lines by typing their label.  [Demo](http://youtu.be/LXiBPvqp_7U)

### DIFF
--- a/src/screenControl.py
+++ b/src/screenControl.py
@@ -31,6 +31,8 @@ SELECT_MODE = 'SELECT'
 COMMAND_MODE = 'COMMAND_MODE'
 X_MODE = 'X_MODE'
 
+lbls = "ABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890~!@#$%^&*()_+<>?{}|;'"
+
 SHORT_NAV_USAGE = '[f|A] selection, [down|j|up|k|space|b] navigation, [enter] open, [c] command mode'
 SHORT_COMMAND_USAGE = 'command examples: | git add | git checkout HEAD~1 -- | mv $F ../here/ |'
 SHORT_COMMAND_PROMPT = 'Type a command below! Files will be appended or replace $F'
@@ -231,7 +233,7 @@ class Controller(object):
 
     def getChromeBoundaries(self):
         (maxy, maxx) = self.stdscr.getmaxyx()
-        minx = CHROME_MIN_X if self.scrollBar.getIsActivated() else 0
+        minx = CHROME_MIN_X if self.scrollBar.getIsActivated() or self.mode == X_MODE else 0
         maxy = self.helperChrome.reduceMaxY(maxy)
         maxx = self.helperChrome.reduceMaxX(maxx)
         # format of (MINX, MINY, MAXX, MAXY)
@@ -335,42 +337,13 @@ class Controller(object):
         self.dirtyHoverIndex()
         self.updateScrollOffset()
 
-    def printXMode(self) :
-        if self.mode == X_MODE :
-            self.mode = X_MODE
-            (maxy, maxx) = self.scrollBar.screenControl.getScreenDimensions()
-            topY = maxy - 2
-            minY = self.scrollBar.getMinY()
-            lbls = "ABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890~!@#$%^&*()_+<>?{}|;'"
-            lblcnt = 0
-            for i in range(minY,topY+1) :
-                self.stdscr.addstr(i, 1, lbls[lblcnt])
-                lblcnt += 1
-
-    def selectXMode(self, key):
-        lblIndex = "ABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890~!@#$%^&*()_+<>?{}|;'".index(key)
-        if self.scrollOffset == 0:
-            lineno = lblIndex + 1
-        else :
-            lineno = lblIndex + 1 + (-1 * self.scrollOffset)
-        l = self.lineObjs[lineno]
-        if hasattr(l,"toggleSelect") :
-            matchIndex = self.lineMatches.index(l)
-            self.dirtyIndexes.append(matchIndex)
-            self.lineMatches[matchIndex].toggleSelect()
-
     def processInput(self, key):
         if key == 'UP' or key == 'k':
             self.moveIndex(-1)
         elif key == 'DOWN' or key == 'j':
             self.moveIndex(1)
         elif key == 'x':
-            if self.mode != X_MODE:
-                self.mode = X_MODE
-                self.printXMode()
-            else:
-                self.mode = SELECT_MODE
-                self.printAll()
+            self.toggleXMode()
         elif key == 'c':
             self.beginEnterCommand()
         elif key == ' ' or key == 'PAGE_DOWN':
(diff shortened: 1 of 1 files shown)


pcottle:
Wow this is awesome!! :heart_eyes: The integration inside of the scrollbar is especially slick, and someone even asked for this internally at FB

Do you mind rebasing on top of the recent changes? It's probably just the scrollbar bug I fixed earlier today

xavierbeynon:
Thanks - done.

src/screenControl.py
```diff
@@ -375,6 +380,8 @@ def processInput(self, key):
             # before exiting the program
             self.getFilesToUse()
             sys.exit(0)
+        elif self.mode == X_MODE and key in lbls:
```
pcottle: we could probably handle this case first and keep the rest of the if/else statement free from checking X_MODE. more future proof probably

src/screenControl.py
```diff
@@ -115,7 +118,7 @@ def outputBottom(self):
         (maxy, maxx) = self.screenControl.getScreenDimensions()
         borderY = maxy - 2
         # first output text since we might throw an exception during border
-        usageStr = SHORT_NAV_USAGE if self.mode == SELECT_MODE else SHORT_COMMAND_USAGE
+        usageStr = SHORT_NAV_USAGE if self.mode == SELECT_MODE or X_MODE else SHORT_COMMAND_USAGE
```
pcottle: probably will move this to a switch statement

src/screenControl.py
```diff
@@ -244,7 +247,7 @@ def getScreenDimensions(self):
 
     def getChromeBoundaries(self):
         (maxy, maxx) = self.stdscr.getmaxyx()
-        minx = CHROME_MIN_X if self.scrollBar.getIsActivated() else 0
+        minx = CHROME_MIN_X if self.scrollBar.getIsActivated() or self.mode == X_MODE else 0
```
pcottle: wonder if we could move this inside `scrollBar.getIsActivated()`...

src/screenControl.py
```diff
@@ -473,8 +480,8 @@ def resetDirty(self):
         self.linesDirty = False
         self.dirtyIndexes = []
 
-    def dirtyHoverIndex(self):
-        self.dirtyIndexes.append(self.hoverIndex)
+    def dirtyHoverIndex(self, idx=-1):
+        self.dirtyIndexes.append(idx if idx > 0 else self.hoverIndex)
```
pcottle: probably should make this a separate function rather than have the optional negative-default peram

src/screenControl.py
```diff
@@ -514,3 +522,25 @@ def moveCursor(self):
...
+        if self.mode == X_MODE:
+            (maxy, _) = self.scrollBar.screenControl.getScreenDimensions()
+            topY = maxy - 2
+            minY = self.scrollBar.getMinY() - (1 if not self.scrollBar.activated else 0)
+            for i in range(minY, topY + 1):
+                self.stdscr.addstr(i, 1, lbls[i - minY])
+
+    def selectXMode(self, key):
+        lineObj = self.lineObjs[lbls.index(key) + int(self.scrollBar.activated) - self.scrollOffset]
+        if hasattr(lineObj, "toggleSelect"):
```
pcottle: probably should filter for the ones that are instances of line match (instead of simple match)

src/screenControl.py
```diff
@@ -514,3 +522,25 @@ def moveCursor(self):
...
+
+    def toggleXMode(self):
+        if self.mode != X_MODE:
+            self.mode = X_MODE
+        else:
+            self.mode = SELECT_MODE
+        self.printAll()
+
+    def printXMode(self):
+        if self.mode == X_MODE:
```
pcottle: nit: if `!=` return

src/screenControl.py
```diff
@@ -514,3 +522,25 @@ def moveCursor(self):
...
+            (maxy, _) = self.scrollBar.screenControl.getScreenDimensions()
+            topY = maxy - 2
+            minY = self.scrollBar.getMinY() - (1 if not self.scrollBar.activated else 0)
+            for i in range(minY, topY + 1):
+                self.stdscr.addstr(i, 1, lbls[i - minY])
```
pcottle: what do you think about making the X mode selectors only appear next to lines that are actual matches?

that would give us more efficient usage of available characters **and** avoid the awkward situation where you hit a key that doesnt correspond to an actual line match

pcottle:
alright cool! i think ill merge this next. curious about your thoughts on:

> what do you think about making the X mode selectors only appear next to lines that are actual matches?

i think that'll be a better experience

src/screenControl.py
mahamudul007: Duoss4462@gmail.com

xavierbeynon:
I avoided that because having a cryptogram (BGFUWXYZ), or a really spaced out ABCDE along the side would be less clear than labeling the lines.  It would also look like the matrix as you scroll.

It could be argued that because x mode is an advanced feature, that it doesn't need to be particularly intuitive.  And you're certainly right - hitting a marked key with no result is awkward.  But IMO that's a symptom of interspersing actionable items among unactionable ones.  I think this would be better solved with color - for example, green for actionable, or perhaps just greying out the unactionable.

pcottle:
> It would also look like the matrix as you scroll.

and thats a bad thing? lol jk

> I think this would be better solved with color - for example, green for actionable, or perhaps just greying out the unactionable.

hrm yeah thats definitely something we could think about. ok cool, i'd be happy to merge this once nits are addressed (if you dont mind, or i can take a crack later this week)

xavierbeynon:
I can work on it this weekend

lastquestion:
since color code is merged, you can have dark grey versus white bold text pretty easily. you can just print decorated text or i would suggest printing with color pairs using `ColorPrinter`. Although probably would need to factor out `setAttributes` into a `getAttributeCode(...)` and `setAttributes` (lines 26-40)

pcottle:
Sweet! Looking forward to it. And yeah with the nice color code, adding decorators to this should be easy

pcottle:
Aight this is a bit out of date so I'm closing out the PR but would definitely be down to see a rebase of this!

src/screenControl.py
xavierbeynon: I think that would be worse.  When X_MODE is on we want to exclude G and A from the keybindings set because they conflict with line letters.  But as of yet we only have an if else block, not a set to subtract from.  There's no case to handle separately.

src/screenControl.py
xavierbeynon: No switch statements in python, so I used the old dictionary select trick.  I agree it makes it cleaner.

src/screenControl.py
xavierbeynon: That would be mixing purposes.  Just because X_MODE is on doesn't mean the scroll bar is too.  I just want to indent in both cases.

src/screenControl.py
xavierbeynon: This got updated to dirtyLine.

src/screenControl.py
xavierbeynon: Agreed

xavierbeynon:
Sorry this is late, but I rebased and went through your nits.  I'll improve the color mode for another time.

xavierbeynon:
I'm not familiar with this - do I have to open another pull request?  The changes to my fork aren't showing up here.

pcottle:
> do I have to open another pull request? The changes to my fork aren't showing up here.

Yeah youll probably have to open up a new PR compared from the tip of your fork to master. I would make sure you're rebased on the latest version of master though -- try setting up your repo to fetch from mine as upstream and merge `upstream/master`

pcottle:
more info here:
https://help.github.com/articles/syncing-a-fork/

PR #48, opened 2015-05-08
If there are more than 3 files, don't use Vim's vertical split

peteryates:
More than 3 files in a vertical split in vim is of little/no use for most people and buffers are more easily navigable.

### DIFF
--- a/src/output.py
+++ b/src/output.py
@@ -141,7 +141,9 @@ def expandPath(filePath):
 def joinEditCommands(partialCommands):
     editor, editor_path = getEditorAndPath()
     if editor == 'vim':
-        if len(partialCommands) > 1:
+        if len(partialCommands) > 3:
+            return editor_path + ' ' + ' '.join(partialCommands)
+        elif len(partialCommands) > 1:
             return editor_path + ' -O ' + ' '.join(partialCommands)
         else:
             return editor_path + ' ' + partialCommands[0]


Mattias-:
3 seems like an arbitrary number.
Personally I'd prefer to have the limit at 2 rather than 3.

peteryates:
It is an arbitrary number; I'm equally as happy with 2

pcottle:
> More than 3 files in a vertical split in vim is of little/no use for most people and buffers are more easily navigable.

I actually disagree -- with setting min col width (`set winwidth=85`) it can be quite manageable to have a bunch of splits but view the current one with >80 cols. I would try it out!

What I don't like about buffers is that its not visually clear from the get-go that you've actually opened all the files, and a lot of people do `:qa` instead of `:n` so they dont even see the other files.

What about splitting files into tabs instead? or throw up a warning?

I actually think we should have some kind of user preferences file to store things like this is maybe the best bet

pcottle:
Closing for now, curious what you guys think about the tabs option. Or setting this as a user preference flag

peteryates:
I'm against vim's tags because I think they're horribly confusing.

The `winwidth` option looks good and isn't too overwhelming.  Unfortunately, though, this solution would require prior knowledge (I'd never heard of it despite having used vim almost-daily for more than twelve years!) and setup in vim, so someone using fpp for the first time could still be overwhelmed by something like this:

screen shot 2015-05-09 at 10 26 23

robenkleene:
Another vote for opening the files in buffers rather than. My expectation is that `fpp` will open files the same way vim would normally open them (e.g., `vim file1.txt file2.txt`, i.e., in buffers). 

The multiple buffer editing feature in vim is designed for this exact case, e.g., it gives a warning when the user quits before addressing all the opened files.

This is a common situation already when composing shell commands with vim (e.g., with `grep` and `xargs`). I don't see any reason `fpp` needs special treatment here.

I also don't see the need for a user preference, if a user wants `fpp` to use tabs instead of buffers, they can already simply set `FPP_EDITOR=vim -O`.

All imo of course, ps for those looking to disable the split behavior setting to be vim followed by a space does the trick: `FPP_EDITOR="vim "`

robenkleene:
Also, with the opening files in tabs behavior, if too many files are opened at once vim will fail with this message: `Too many "+command", "-c command" or "--cmd command" arguments`

PR #52, opened 2015-05-08
fix help text

rmetzler:
- there was a typo: "HEAD~10" vs "HEAD~1". I chose the latter, because I think it is used more often.
- etc is a comment in a shell command, that's why I replaced // with #

### DIFF
--- a/src/processInput.py
+++ b/src/processInput.py
@@ -64,13 +64,13 @@ enter before it is executed, so all you have
 to do is type the prefix. Some examples:
 
     * git add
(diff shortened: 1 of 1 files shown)


pcottle:
Thanks a bunch! The `# etc` comment is definitely an improvement

sometimes the `~10` shows a more interesting output than just `~1`, but i guess we are being quite assumptive about the history of your repo (assuming it has more than 9 commits). so im ok with that update -- not a huge leap forward but hopefully our ascii demos do enough talking for us

PR #53, opened 2015-05-08
ANSI color support

lastquestion:
~~This is a shitty first pass at colors that I hacked together because I couldn't sleep.~~ Now better!

Turns out `ncurses` doesn't really support (AFAIK, someone please correct me...) setting foreground/background color combinations directly, so I cache and create `color_pair` as I go along. Most terms have at least 32 and mine on OSX seems to have 256.

I'm also totally not a python developer, so the code here is probably quite not idomatic.

Screencap because I don't know how to do that cool video thing:
the command was `git diff --color | fpp`
screen shot 2015-05-08 at 3 30 50 am

Things to improve on
* Make the chrome use colors instead of hardcoding `color_pair(0)`.

### DIFF
--- a/src/colorPrinter.py
+++ b/src/colorPrinter.py
@@ -14,6 +14,9 @@ class ColorPrinter(object):
     def __init__(self, screen):
         self.colors = {}
         self.colors[(0,0)] = 0 #0,0 = white on black is hardcoded
+        #in general, we want to use -1,-1 for most "normal" text printing
+        self.colors[(-1,-1)] = 1
+        curses.init_pair(1, -1, -1)
         self.screen = screen
 
 
@@ -38,3 +41,9 @@ class ColorPrinter(object):
 
     def restoreAttributes(self):
         self.screen.attrset(self.currentAttributes)
+
+    def addstr(self, y, x, text, attr=None):
+        if attr is None:
+            attr = curses.color_pair(1)
+
+        self.screen.addstr(y, x, text, attr)
--- a/src/colorPrinter.py
+++ b/src/colorPrinter.py
@@ -18,13 +18,16 @@ class ColorPrinter(object):
 
 
     def setAttributes(self, foreColor, backColor, other):
+        colorIndex = -1
         colorPair = (foreColor, backColor)
         if not colorPair in self.colors:
             newIndex = len(self.colors)
-            curses.init_pair(newIndex, foreColor, backColor)
-            self.colors[colorPair] = newIndex
-
-        colorIndex = self.colors[colorPair]
+            if newIndex < curses.COLOR_PAIRS:
+                curses.init_pair(newIndex, foreColor, backColor)
+                self.colors[colorPair] = newIndex
+                colorIndex = newIndex
+        else:
+            colorIndex = self.colors[colorPair]
 
         attr = curses.color_pair(colorIndex)
         attr = attr | other
--- a/src/colorPrinter.py
+++ b/src/colorPrinter.py
@@ -1,0 +1,37 @@
+# Copyright (c) 2015-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+#
+# @nolint
+import curses
+
+class ColorPrinter(object):
+    """A thin wrapper over screens in ncurses that caches colors and
+    attribute state"""
+    def __init__(self, screen):
+        self.colors = {}
+        self.colors[(0,0)] = 0 #0,0 = white on black is hardcoded
+        self.screen = screen
+
+
+    def setAttributes(self, foreColor, backColor, other):
+        colorPair = (foreColor, backColor)
+        if not colorPair in self.colors:
+            newIndex = len(self.colors)
+            curses.init_pair(newIndex, foreColor, backColor)
+            self.colors[colorPair] = newIndex
+
+        colorIndex = self.colors[colorPair]
+
+        attr = curses.color_pair(colorIndex)
+        attr = attr | other
+
+        self.currentAttributes = attr
+
+        self.screen.attrset(self.currentAttributes)
+
+    def restoreAttributes(self):
+        self.screen.attrset(self.currentAttributes)
--- a/src/colorPrinter.py
+++ b/src/colorPrinter.py
@@ -5,7 +5,6 @@
 # LICENSE file in the root directory of this source tree. An additional grant
 # of patent rights can be found in the PATENTS file in the same directory.
 #
-# @nolint
 import curses
 
 class ColorPrinter(object):
--- a/src/format.py
+++ b/src/format.py
@@ -8,33 +8,38 @@
 from __future__ import print_function
 
 import curses
-
 import parse
-
+import copy
+from formattedText import FormattedText
 
 class SimpleLine(object):
 
-    def __init__(self, line, index):
-        self.line = line
+    def __init__(self, formattedLine, index):
+        self.formattedLine = formattedLine
         self.index = index
 
     def printOut(self):
         print(str(self))
 
-    def output(self, stdscr):
+    def output(self, printer):
         (minx, miny, maxx, maxy) = self.controller.getChromeBoundaries()
-        maxLen = maxx - minx
+        maxLen = min(maxx - minx, len(str(self)))
         y = miny + self.index + self.controller.getScrollOffset()
+
+        if (y < miny or y > maxy):
+            # wont be displayed!
+            return
+
         try:
-            stdscr.addstr(y, minx, str(self)[0:maxLen])
-        except curses.error:
+            self.formattedLine.printText(y, minx, printer, maxLen)
+        except curses.error as e:
             pass
 
     def setController(self, controller):
         self.controller = controller
 
     def __str__(self):
-        return self.line
+        return str(self.formattedLine)
 
     def isSimple(self):
         return True
@@ -42,8 +47,8 @@ class SimpleLine(object):
 
 class LineMatch(object):
 
-    def __init__(self, line, result, index):
-        self.line = line
+    def __init__(self, formattedLine, result, index):
+        self.formattedLine = formattedLine
         self.index = index
 
         (file, num, matches) = result
@@ -51,6 +56,8 @@ class LineMatch(object):
         self.originalFile = file
         self.file = parse.prependDir(file)
         self.num = num
+
+        line = str(self.formattedLine)
         # save a bunch of stuff so we can
         # pickle
         self.start = matches.start()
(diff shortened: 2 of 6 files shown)


pcottle:
Very cool! This is a pretty beefy diff so it will take a while to look over, but just some high level comments:

  * how does this work on light terminal themes or terminals with different default colors? we just added support for using default colors so I want to make sure this looks ok
  * this basically just respects the color formatting of the input right, not adds any custom coloring? im wondering if theres an easier way to translate terminal color codes to ncurses color codes...

lastquestion:
@pcottle yeah it's a big diff and I also don't like a lot of how I hacked it. I realized this morning that the printing for matched lines can be much simpler now that we can actually inject the attributes we want (underscore, background color) directly into the formatted line; we wouldn't need to print the before(), then middle, then after(). Also, there's some complexity around keeping the `formatted_line` versus the stripped line and doing matching and keeping the indices in order, but if I'm super awake and actually white board the cases out I think I can rewrite it to use regex and split out the lines into formatted regions.

* Good point: it should use -1 for the background color so it uses the default. Did someone just add `use_default_colors()`? I rebased and merged pretty quickly and noticed that but didn't `git blame` to see how that cropped up; when I was hacking I don't remember it there

* yes - it should take the input color formatting, apply the right attrs, and only override when the matched line requires it because you're hovering over the proposed link. and i was hella surprised that there wasn't something already out there to do this... I searched but couldn't find anything except a super old post in a google group about it.

I am going to work on it this weekend a little bit on the plane if I can.

pcottle:
> Did someone just add use_default_colors()?

yep! it was @frantic, great addition imo

>  and i was hella surprised that there wasn't something already out there to do this... I searched but couldn't find anything except a super old post in a google group about it.

ah ok, that's too bad. yeah i dont think curses UIs in python are thattttt common to begin with.

We should be concerned with performance though, since with super long inputs I have actually noticed pathpicker starting to slow down / chug during the parsing and scrolling steps. I did some performance work earlier but want to make sure we dont ship any big regressions

lastquestion:
@pcottle this is a much, much better implementation. 

This implementation splits on regex, is much cleaner, simplifies the code for doing decorations on the matches, and works on default light color terms as well. 

I think this is reviewable. I would like to see unit tests and I would write them. Do you have strong preferences on testing infra on python? As I don't usually write python I have none.

pcottle:
>  I would like to see unit tests and I would write them. Do you have strong preferences on testing infra on python? As I don't usually write python I have none.

yeah unfortunately we only have unit testing on the file parsing / regexing / prepending part :-/ I've googled a fair amount and haven't seen anything on how to possibly test curses-based UIs. I would even be down for just a generated file of expected vs actual output but even that's hard to produce I believe.

If you find anything on this let me know -- would be big improvement to the manual UI testing we have to do today

src/colorPrinter.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) 2015-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+#
+# @nolint
```
pcottle: this is actually for FB infra, I can remove it in a followup though

src/format.py
```diff
@@ -9,33 +9,38 @@
...
+
+        if (y < miny or y > maxy):
+            # wont be displayed!
+            return
+
         try:
-            stdscr.addstr(y, minx, str(self)[0:maxLen])
-        except curses.error:
+            self.formattedLine.printText(y, minx, printer, maxLen)
+        except curses.error as e:
```
pcottle: might have to remove this since its different in python 3.0 and we dont use the exception anyways

src/colorPrinter.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def setAttributes(self, foreColor, backColor, other):
+        colorIndex = -1
+        colorPair = (foreColor, backColor)
+        if not colorPair in self.colors:
+            newIndex = len(self.colors)
+            if newIndex < curses.COLOR_PAIRS:
+                curses.init_pair(newIndex, foreColor, backColor)
+                self.colors[colorPair] = newIndex
+                colorIndex = newIndex
```
pcottle: dont think you really need this line if you drop the else case back our

src/colorPrinter.py
```diff
@@ -0,0 +1,40 @@
...
+        else:
+            colorIndex = self.colors[colorPair]
+
+        attr = curses.color_pair(colorIndex)
+        attr = attr | other
+
+        self.currentAttributes = attr
```
pcottle: can we initialize this to something in `__init__`?

src/format.py
```diff
@@ -121,42 +128,47 @@ def getSelected(self):
...
-        curses.init_pair(2, curses.COLOR_WHITE, curses.COLOR_BLUE)
-        curses.init_pair(3, curses.COLOR_WHITE, curses.COLOR_GREEN)
+        return (self.getBefore() + '||' + self.getMatch()
+                + '||' + self.getAfter() + '||' +
+                str(self.num))
 
+    def getDecoratorFormatting(self):
         if self.hovered and self.selected:
-            return curses.color_pair(3)
+            attributes = (curses.COLOR_WHITE, curses.COLOR_RED, 0)
```
pcottle: oooh nice, this is way more straight-forward than the hacky init_pair stuff I was doing :tada:

src/format.py
```diff
@@ -121,42 +128,47 @@ def getSelected(self):
...
 
-    def output(self, stdscr):
+    def output(self, printer):
         decorator = self.getDecorator()
-        before = self.getBefore()
-        after = self.getAfter()
-        middle = ''.join([decorator, self.getMatch()])
+        # inject the decorator and decorated style
+        # into the formatted text and then simply print it out
+        text = copy.deepcopy(self.formattedLine)
```
pcottle: is deep copy performant enough to do this on every output?

src/formattedText.py
```diff
@@ -0,0 +1,127 @@
...
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+#
+# @nolint
+
+import re
+import curses
+
+class FormattedText(object):
+    """A piece of ANSI escape formatted text which responds
```
pcottle: _like_

src/formattedText.py
```diff
@@ -0,0 +1,127 @@
...
+    @classmethod
+    def parseFormatting(cls, formatting):
+        """Parse ANSI formatting; the formatting passed in should be
+        stripped of the control characters and ending character"""
+        fore = -1 #-1 default means "use default", not "use white/black"
+        back = -1
+        other = 0
+        intValues = [int(value) for value in formatting.split(';') if value]
+        for code in intValues:
+            if code >= 30 and code <= 39:
```
pcottle: should these `30` and `40`'s be constants in the class? then we could at least explain what they mean and there would be less magic numbers here

src/processInput.py
```diff
@@ -135,15 +136,16 @@ def getLineObjs():
     lineObjs = {}
     for index, line in enumerate(inputLines):
         line = line.replace('\t', '    ')
-        line = re.sub(r'\x1b[^mK]*(m|K)', '', line)
-        result = parse.matchLine(line)
+        formattedLine = FormattedText(line)
+        result = parse.matchLine(str(formattedLine))
```
pcottle: _like_

src/screenControl.py
```diff
@@ -105,9 +106,9 @@ def outputSide(self):
         if self.mode == COMMAND_MODE:
             usageLines = processInput.USAGE_COMMAND.split('\n')
         for index, usageLine in enumerate(usageLines):
-            self.stdscr.addstr(self.getMinY() + index, borderX + 2, usageLine)
+            self.stdscr.addstr(self.getMinY() + index, borderX + 2, usageLine, curses.color_pair(0))
```
pcottle: the `curses.color_pair(0)` here is a lot of copy-pasta, could we consolidate that somehow? maybe a `self.addStrStandard` or something that uses the default formatting?

pcottle:
alright I cruised through and left some comments, very exciting! its super cool to see the color show up and makes the current version of fpp seem super basic (that it doesnt have color support). now if only git diff would always include color when piping to our program...

do you mind addressing the nits? most of it is just comments or class constants to explain some of the more confusing parts

lastquestion:
@pcottle I think in general most pipe tools seem to support color so I personally just have it turned on all the time but I'm not sure... probably we can add something at the readme level mentioning color. 

```
You can also set it to always to ignore the difference between terminals and pipes. You’ll rarely want this; in most scenarios, if you want color codes in your redirected output, you can instead pass a --color flag to the Git command to force it to use color codes. The default setting is almost always what you’ll want.
```

I will just update the code w/ your comments and rebase

src/colorPrinter.py
lastquestion: Cool. I just copied it over

src/format.py
lastquestion: Yeah, so interesting thing: I was getting the exception before because `SimpleLine` didn't have the guard around y height being too big or too small. In current `master` we are actually throwing and catching QUITE A LOT - which might have some perf ramnifications on super big files.

After adding the guard check I didn't get any more curses exceptions. Personally I think we should allow those to propogate out and if you agree I will remove on both prints

src/colorPrinter.py
lastquestion: do you mean `currentAttributes` or the `colors` dict? `colors` is a cache of the set of colors as requested during printing. It could be preinitialized to the 7x7 default colors array, but I don't know if it would improve anything because anything outside that set would still need the code to add a new color pair.

for `currentAttributes` .. really we shouldn't need it but `curses` doesn't expose `attr_get`. Unless it does. I grepped but didn't see anything. During __init__ we know the attribute set to be (0,0) so we could preinitialize to 0. Is that wat you mean?

src/format.py
lastquestion: Actually I don't know. I suppose we don't really need to do a copy at all or inject.
Actually, it's much simpler and I will reimplement it this way:

1. Break the line at the beginning and end of the match in __init__, therefore it will be saved by `pickle`, this will happen during `processInput` once only, and actually just save two lines instead of one. Use existing `breakat` and remove the `replace` function.
2. When outputting, output each line segment separately, and also, cache the current `formattedText` representing the decorator. 

It removes the alloc on each output and is also cleaner and removes the need for `replace` functionality. Feels good

src/formattedText.py
lastquestion: Yeah :+1: lets do that

src/screenControl.py
lastquestion: Yeah, let's do that. Actually we can later make our chrome colored which would be sweet. For later.

pcottle:
> You can also set it to always to ignore the difference between terminals and pipes. You’ll rarely want this; in most scenarios,

ah interesting!

> I will just update the code w/ your comments and rebase

sweet, looking forward to it :joy:

lastquestion:
Wow, the amount of code change going in is large...rebasing every day makes me wanna get this in now.

OK, so I updated per comments, and now we cache the decorated matches actually so I think perf ought to be pretty good. I manually tested on light / dark chrome and ran the test suite. all passed

pcottle:
:tada:

src/colorPrinter.py
pcottle: done

src/colorPrinter.py
pcottle: yeah I basically meant just so we know that this instance variable is used elsewhere in code -- aka you can see there are 3 instance variables from looking at init

pcottle:
Alright thanks a bunch @lastquestion! It's pretty badass to have the color show up in fpp :D also glad you split the color printing out into a new class and kept the logic fairly self contained. this will make it in the next release for sure (maybe EOW)

lastquestion:
hey @pcottle it would be much easier if README or CONTRIB mentioned that pep8 is required so you don't have to keep on fixing other PR's. It's easy to run and I just didn't know we needed to run it.

pcottle:
@lastquestion i'll be part of the travis CI run shortly!

pcottle:

Issue #49, opened 2015-05-08
Support the $VISUAL env variable if set


jacobk:
http://unix.stackexchange.com/questions/4859/visual-vs-editor-whats-the-difference

pcottle:
@jacobk does using `FPP_EDITOR` solve your use case? or would you like us to check `$VISUAL` first?

pcottle:
:tada:

PR #58, opened 2015-05-08
Allow a different editor than $EDITOR to be used.

deefour:
Opted against `$VISUAL` as mentioned in #49 in favor of `$FPP_EDITOR` for more flexibility.

For example, `$EDITOR` is set to vim and `$VISUAL` is set to `subl` for me, but I tend to prefer paths launched from PathPicker to open in TextMate. Now I can set `$FPP_EDITOR` to `mate`

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,10 @@ to `/usr/local/bin/` which is assumed to be in the current
 * `ln -s ./fpp /usr/local/bin/fpp`
 * `fpp --help # should work!`
 
+## Configuration
+
+The `$FPP_EDITOR` environment variable can be set to tell PathPicker which editor to open paths with. The `$EDITOR` environment variable is used by default.
+
 ## Advanced Functionality
 
 As mentioned above, PathPicker allows you to also execute arbitrary commands with the specified files.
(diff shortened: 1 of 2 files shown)


davvid:
I implemented this same exact feature (with this same exact env. variable name) before checking here and realizing that you did the same.  :+1: for this feature, this would be very helpful.

pcottle:
Maybe we should add this to `USAGE` as well? i figure it should output during `fpp --help` as well as in the readme

actually what do you think about putting it in only `--help`? Might be a better place for it anyways

pcottle:
Yeah going to move this into the `--help` output and maybe lower in the README :P

pcottle:
```

== Configuration == 


PathPicker offers a bit of configuration currently with more to come
in the future.

~ Editor ~

The $FPP_EDITOR environment variable can be set to tell PathPicker
which editor to open the selected files with. If that variable
is not set, $EDITOR is used the next fallback, with "vim" as a last resort.


That's a fairly in-depth overview of Facebook PathPicker.
We also provide help along the way as you
use the app, so don't worry and jump on in!

********************************************************************************
```

deefour:
@pcottle thanks! :beers:

Issue #51, opened 2015-05-08
Allow use of shell aliases


raine:
I have a ton of git aliases and would like to use them with command mode. Otherwise the mode is practically worthless.

`fbPager.sh: line 4: gco: command not found`

joeguilmette:
a big +1 from me

pcottle:
Alright I heard back about my approach in #59 so lets go with that! My only worry is that we source the entire `.bashrc` which might have side-effects. I actually use my bashrc to log onto my development server for facebook (automatically) so its kind of an aggressive script to source, but hopefully my situation is not super common.

```
[pcottle:~/Dropbox (Facebook)/wip/PathPicker:master]$ git show --stat | ./fpp
getting input
executing command:
ga  '/Users/pcottle/Dropbox (Facebook)/wip/PathPicker/src/output.py'
```
(`ga` is aliased to git add for me)

should work 100% for bash, and hoping zsh is the same way. can anyone test?

langtianlang:
@pcottle Using 0.5.6. Zsh aliases did not work for me

```
➜  code/stuff git:(master) ✗ gs | fpp
getting input
executing command:
ga '/Users/llang/code/stuff/thing.rb'
/Users/llang/.fpp/.fpp.sh: line 4: ga: command not found
➜  code/stuff git:(master) ✗ brew upgrade fpp
Error: fpp 0.5.6 already installed
➜  code/upclookup git:(master) ✗ ga
Nothing specified, nothing added.
Maybe you wanted to say 'git add .'?
```

pcottle:
@langtianlang youll have to install from master, since 0d12fef0bcfd3e2c4ac2e4e72af7f35041931c7e is not in fpp `0.5.6`:

```
# excerpt from 0.5.6's output.py
def getAliases():
    try:
        lines = open(os.path.expanduser(BASH_RC), 'r').readlines()
    except IOError:
```

either follow the manual install instructions in the README or wait till tonight when I submit the new release :)

Issue #32, opened 2015-05-07
Arbitrary command mode doesn't have aliases or shell functions


SeanA208:
When executing a command in arbitrary command mode, fpp is unaware of any aliases or shell functions I may have defined in my .zshrc (I'm using zsh but I've tested this in bash as well). 

For example I have an alias `gco="git checkout"`and running `gco` in the arbitrary command mode gives:

```.fbPager.sh: line 4: gco: command not found```

pcottle:
Yeah @SeanA208  we actually dont get any of the aliases that are already loaded in the shell script since we execute in another shell environment :*( So for bash we have to **manually** expand the aliases:

https://github.com/facebook/PathPicker/blob/master/src/output.py#L55

However no reason why we can't make a fallback...

pcottle:
Alright this is a bit hacky but e0604190d46c03b61869607fc2cc1b725b72a800 should work. Can you try that? I don't have zsh installed unfortunately :-/

(obviously this isnt reflected in prod so youll have to make those changes manually to your brew installation)

thanks @SeanA208 for the report!

PR #59, opened 2015-05-08
Make shell expand aliases (work in progress)

frantic:
Instead of parsing and extracting shell aliases from `.bashrc` and `.zshrc`, would be nice to find a way to convince shell to do this for us.

This is first iteration - it works with zsh but doesn't work with bash. My test env is as following:

```bash
# .zshrc
source ~/.alias
```

```bash
# .bashrc
source ~/.alias
```

```bash
alias ll='ls -lah'
```

Might resolve issues like #51, #32

### DIFF
--- a/fpp
+++ b/fpp
@@ -48,5 +48,10 @@ $PYTHONCMD "$BASEDIR/src/processInput.py"
 # now choose input and...
 exec 0<&-
 $PYTHONCMD "$BASEDIR/src/choose.py" < /dev/tty
-# execute the output bash script
-sh ~/.fbPager.sh < /dev/tty
+
+# Try to guess which shell this script was launched in
+SHELL="$(ps -o comm= $PPID)"
+echo running in $SHELL
+# Run shell in login interactive mode, this forces it
+# to load .zshrc/.bashrc/etc that may contain aliases
+$SHELL -i ~/.fbPager.sh < /dev/tty
--- a/src/output.py
+++ b/src/output.py
@@ -14,8 +14,6 @@ import logger
 
 OUTPUT_FILE = '~/.fbPager.sh'
 SELECTION_PICKLE = '~/.fbPager.selection.pickle'
-BASH_RC = '~/.bashrc'
-ZSH_RC = '~/.zshrc'
 DEBUG = '~/.fbPager.debug.text'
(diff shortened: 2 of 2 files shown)


shaneharter:
If the goal is just to support aliases, why not just remove the code that scans for aliases in .bashrc (etc) and instead just use the output of /usr/bin/alias ?

pcottle:
@shaneharter I don't see that file having any of my aliases defined in my `.bashrc`:

```
[pcottle:~:]$ cat /usr/bin/alias 
#!/bin/sh
# $FreeBSD: src/usr.bin/alias/generic.sh,v 1.2 2005/10/24 22:32:19 cperciva Exp $
# This file is in the public domain.
builtin `echo ${0##*/} | tr \[:upper:] \[:lower:]` ${1+"$@"}
[pcottle:~:]$ type -a go
go is aliased to `git checkout'
```

I think most people load their aliases through a bashrc/zshrc, so that might not work super well?

pcottle:
@frantic My crack at this is over in 0ae54808846d058f2e89a08501dab61a628f1434. Basically it adds the expand aliases option before sourcing (conditionally) a list of files. Output looks like:

```
shopt -s expand_aliases
if [ -f ~/.zshrc ]; then
  source ~/.zshrc
fi
if [ -f ~/.bashrc ]; then
  source ~/.bashrc
fi
if [ -f ~/.bash_profile ]; then
  source ~/.bash_profile
fi
if [ -f ~/.bash_aliases ]; then
  source ~/.bash_aliases
fi
echo "executing command:"
echo "go  './banana.py'"
go  './banana.py'
```

Works well for bash and you get real sourcing. Looks like the zsh equivalent is:
http://stackoverflow.com/questions/23258413/expand-aliases-in-non-interactive-shells
`setopt aliases` ?

What do you guys think about doing that? Seems less hacky that introspecting the process ID

(btdubs I guess I shouldn't pollute the repos paths right? ill try to do PR's in the future)

frantic:
Cool, makes sense!

pcottle:
@frantic now that this is merged can you try in zsh land? I think we will need to guard the `setop` and `shopt` commands behind some kind if statement (or just output their error to dev null)

PR #61, opened 2015-05-08
fpp: add a --keep-open flag

davvid:
Add a flag that allows you to keep fpp running after running a command
or editing a file.

Run the current command but do not exit fpp when the -k|--keep-open flag
is specified on the command-line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,10 @@ in the middle of your command, you can use the `$F` token instead, like:
 
 `cat $F | wc -l`
 
+By default, `fpp` exits once you execute a command.
+Use `fpp --keep-open` (or `fpp -k`) to keep your `fpp` session open
+after running a command.
+
 ## How PathPicker works
 PathPicker is a combination of a bash script and some small Python modules.
 It essentially has three steps:
--- a/fpp
+++ b/fpp
@@ -47,6 +47,6 @@ done
 $PYTHONCMD "$BASEDIR/src/processInput.py"
 # now choose input and...
 exec 0<&-
-$PYTHONCMD "$BASEDIR/src/choose.py" < /dev/tty
+$PYTHONCMD "$BASEDIR/src/choose.py" "$@" < /dev/tty
 # execute the output bash script
 sh ~/.fbPager.sh < /dev/tty
--- a/src/choose.py
+++ b/src/choose.py
@@ -6,6 +6,7 @@
 # of patent rights can be found in the PATENTS file in the same directory.
 #
 # @nolint
+import argparse
 import curses
 import pickle
 import sys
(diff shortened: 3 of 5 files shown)


lastquestion:
:+1: this is really nice and I have a parallel PR that's worse then your implementation

pcottle:
Thanks for the PR @davvid ! Curious about the workflow here -- the only tricky thing is now the execution of the shell script is split between python and the bash script (since now we have the `os.system('sh ~/.fbPager.sh; rm -f ~/.fbPager.sh'`) so it complicates things like #59 where we need to decorate the bash script with a few headers

Also does this work for editing files? We would need to connect to `dev/tty` in that case right?

I almost wonder if we could implement this instead as pure bash, something like

```
$PYTHONCMD "$BASEDIR/src/processInput.py"
function runChooser {
  exec 0<&-
  chose.py // etc
  sh ~/.fbPager.sh < /dev/tty
}

if $1 == '--keep-open'; then
  while [true]:
    runProgram()
else
  runProgram()
fi
```

since we save the pickled selection and the files, this should give almost the exact same behavior without introducing the code complexity. what do you think?

davvid:
I did test it with editing files (that was my use case: vim and gvim), so that did work, but as you said, it made the code a bit complex.

If no one beats me to it then I'll try to rework it with your suggestions in mind.  Thanks for the review

pcottle:
> If no one beats me to it then I'll try to rework it with your suggestions in mind. Thanks for the review

no problem @davvid, thanks for the interest in the project! After seeing how many changes have (recently) gone into the python stack and how many are planned (like moving the location of the pickled files), it'd be nice to keep that as simple as possible.

pcottle:
(lemme know if you cant get around to the refactor and I can take a stab and close out this PR)

davvid:
thanks y'all. please feel free to take a stab at it -- it'll probably be
about a week or two before I'll have a chance to take a look.

pcottle:
yeah sure lemme take a stab at this

pcottle:
Boom!

```
[pcottle:~/Dropbox (Facebook)/wip/PathPicker:master]$ ./fpp -ko
Using old result...
nothing to do!
Using old result...
nothing to do!
Using old result...
nothing to do!
Using old result...
executing command:
cat '/Users/pcottle/Dropbox (Facebook)/wip/PathPicker/scripts/buildAndTest.sh' '/Users/pcottle/Dropbox (Facebook)/wip/PathPicker/src/format.py' | wc -l
     199
Using old result...
2 files to edit
Using old result...
nothing to do!
Using old result...
```

through the magic of stack overflow and ridiculous bash programming we have this down :tada: :tada:

PR #62, opened 2015-05-08
First step to make the code more pythonic

dduong42:
In this PR:
- Remove some getters/setters.
- Remove change some `if len(l)` into `if l`
- Change the unused variables into `_` when unpacking.
- Remove some unused code.
- Add the `ignore_curse_errors` context manager (avoid the `try`/`pass`)
- Use `values` with a dictionary when the key is not needed.
- Remove the `controller` attribute from the `SimpleLine` and the `LineMatch`. (Pass it to `output` instead).
- ...

To sum up, this a first step to make the code a little bit more pythonic. There is some work left, but I didn't want to make a PR with too many commits. The `format.py` file should be a little less "java".

### DIFF
--- a/src/choose.py
+++ b/src/choose.py
@@ -13,6 +13,7 @@ import pickle
 import sys
 import os
 
+from format import LineMatch
 import output
 import screenControl
 import logger
@@ -33,7 +34,7 @@ def getLineObjs():
     filePath = os.path.expanduser(PICKLE_FILE)
     lineObjs = pickle.load(open(filePath))
     matches = [lineObj for lineObj in lineObjs.values()
-               if not lineObj.isSimple()]
+               if isinstance(lineObj, LineMatch)]
     logger.addEvent('total_num_files', len(lineObjs))
 
     selectionPath = os.path.expanduser(SELECTION_PICKLE)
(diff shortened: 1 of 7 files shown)


pcottle:
Wow this is huge @dduong42 ! I don't write much python at all so a lot of the code is nowhere near pythonic (as you can probably tell)

Before I take a closer look at this, the function decorators are supported the same in python 3.0 right? We are close to getting python 3.0 support so don't want to regressin that sense

dduong42:
The changes won't affect the python 3 compatibility :).

zkanda:
I suggest breaking this into multiple pr's, being more pythonic is very subjective.

But I agree, using `property` decorator is very pythonic. :)

mahamudul007:
Duoss4462@gmail.com

pcottle:
Yeah do you mind breaking this up into multiple PR's? Maybe on per file or class. We have a ton of incoming PR's so I think that'll be easier on everyone, since this will continue to get rebased on top of and go stale

all about more succinct python though :heart_eyes:

pcottle:
(Closing for now, lets move to multiple PR's)

PR #64, opened 2015-05-08
Remove Facebook-specific examples

Daniel15:


pcottle:
Ah my bad @Daniel15 !! @sahrens pointed this out in a comment thread and i immediately committed it without checking PR's, so I got to this before you :cry: definitely want to get another FB contributor though so try to find something else! maybe the sudo app-get support?

Daniel15:
Haha that's okay, no worries :+1:

PR #71, opened 2015-05-09
Fix pickle issue in Python3

zkanda:
According to the docs, we need to read and write in `bytes`:

https://docs.python.org/3.4/library/pickle.html#data-stream-format
Protocol version 3 was added in Python 3.0. It has explicit support for bytes objects and cannot be unpickled by Python 2.x. This is the default protocol, and the recommended protocol when compatibility with other Python 3 versions is required.

### DIFF
--- a/src/choose.py
+++ b/src/choose.py
@@ -31,7 +31,7 @@ def doProgram(stdscr):
 
 def getLineObjs():
     filePath = os.path.expanduser(PICKLE_FILE)
-    lineObjs = pickle.load(open(filePath))
+    lineObjs = pickle.load(open(filePath, 'rb'))
     matches = [lineObj for i, lineObj in lineObjs.items()
                if not lineObj.isSimple()]
     logger.addEvent('total_num_files', len(lineObjs.items()))
--- a/src/output.py
+++ b/src/output.py
@@ -82,7 +82,7 @@ def debug(*args):
 def outputSelection(lineObjs):
     filePath = os.path.expanduser(SELECTION_PICKLE)
     indices = [l.index for l in lineObjs]
-    pickle.dump(indices, open(filePath, 'w'))
+    pickle.dump(indices, open(filePath, 'wb'))
 
 
 def getEditorAndPath():
--- a/src/processInput.py
+++ b/src/processInput.py
@@ -98,7 +98,7 @@ Which format to:
 '''
 
 USAGE_CONFIGURATION = '''
-== Configuration == 
+== Configuration ==
 
 
 PathPicker offers a bit of configuration currently with more to come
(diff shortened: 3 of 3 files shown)


zkanda:
I think we can already revert the changes made here: https://github.com/facebook/PathPicker/blob/master/fpp#L32

pcottle:
one thing to note is that this makes our previous-selection-loading behavior not be backwards compatible:

```
[pcottle:~/Dropbox (Facebook)/wip/PathPicker:]$ ./fpp
Using old result...
  File "/Users/pcottle/Dropbox (Facebook)/wip/PathPicker/src/choose.py", line 58, in <module>
    curses.wrapper(doProgram)
  File "/usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework/Versions/3.4/lib/python3.4/curses/__init__.py", line 94, in wrapper
    return func(stdscr, *args, **kwds)
  File "/Users/pcottle/Dropbox (Facebook)/wip/PathPicker/src/choose.py", line 27, in doProgram
    lineObjs = getLineObjs()
  File "/Users/pcottle/Dropbox (Facebook)/wip/PathPicker/src/choose.py", line 34, in getLineObjs
    lineObjs = pickle.load(open(filePath, 'rb'))
ImportError: bad magic number in 'formattedText': b'\x03\xf3\r\n'
```

so ill add something for this:
screen shot 2015-05-10 at 10 21 05 am

src/choose.py
```diff
@@ -31,7 +31,7 @@ def doProgram(stdscr):
 
 def getLineObjs():
     filePath = os.path.expanduser(PICKLE_FILE)
-    lineObjs = pickle.load(open(filePath))
+    lineObjs = pickle.load(open(filePath, 'rb'))
     matches = [lineObj for i, lineObj in lineObjs.items()
                if not lineObj.isSimple()]
     logger.addEvent('total_num_files', len(lineObjs.items()))
```
pcottle: i think you also need to change the `pickle.load(open(selectionPath)` down below (github wont let me comment that far down)

pcottle:
Alright this is the last python3 error I'm getting. selecting files to edit works fine but in command mode:

```
  File "/Users/pcottle/Dropbox (Facebook)/wip/PathPicker/src/output.py", line 176, in composeFileCommand
    if '$F' in command:
TypeError: 'str' does not support the buffer interface
```

pcottle:
Boom! I did a fairly wide audit and I think I hit up all code branches, and I think thats all we need for python3.0 support! :tada: :fireworks: thanks @zkanda for doing a big chunk of this

i removed the changes you mentioned, so now python3.0 peeps will be running the real deal

PR #73, opened 2015-05-09
Update README.md

sametmax:
Use ls on $(readlink -f ./fpp) instead of on relative path, which can lead to the "Too many levels of symbolic links" error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Here we make a symbolic link from the bash script in the repo
 to `/usr/local/bin/` which is assumed to be in the current
 `$PATH`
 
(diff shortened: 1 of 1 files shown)


sametmax:
I won't do that. Good bye.

pcottle:
Sorry, the CLA is required for us to merge this :-/ Also I'm not a fan of needing the `sudo` part either. Also keep in mind that `readlink -f` is not supported on osx!

```
[pcottle:~/Dropbox (Facebook)/wip/PathPicker:]$ readlink -f
readlink: illegal option -- f
```

so dont think this can go in

Issue #17, opened 2015-05-07
PathPicker pollutes user home directory


vitaut:
PathPicker creates multiple files in the user home directory:

```
$ ls .fb*
.fbPager.log  .fbPager.pickle  .fbPager.selection.pickle  .fbPager.sh
```

Normally such files are stored under `$HOME/.local/share` or `$HOME/.config` directory on Linux. Please see [XDG Base Directory Specification](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) for details.

pcottle:
Yeah I have been meaning to get around to this -- totally can as well. a few questions for you @vitaut:

* is the current user always guaranteed to have write capabilities into `$HOME/.local/share`? one of the nice things about the home directory is not having to worry about user permissions
* is it guaranteed to exist? or will we have to `mkdir -p` beforehand?

I'd love to resolve this though, and it shouldnt be hard since these are all constants in the code

vitaut:
Although all systems I've been working on have writable `$HOME/.local/share`, I can't say whether this directory is guaranteed to exist, so it's probably safer to check and create if it doesn't.

On a closer inspection though`.fbPager*` files look more like temporary files (are they?) rather than some user-specific settings and could probably be stored in `/tmp`.

pcottle:
yeah they are all temporary -- the only one that persists between sessions is the selection pickle and if thats gone its nbd.

same question -- is `/tmp` always guaranteed to be writeable? or on average more guaranteed to be writeable and exist compared to a home dir?

vitaut:
I think it's safe to assume that `/tmp` is writable.

pcottle:
Alright -- this is coming tomorrow!

we'll be sharing the same pickle selection across users on the same machine, so maybe we should prepend by the username? `pwd.getpwuid(os.getuid()).pw_name`

zkanda:
I think it's safe to assume that everything under the users directory is owned by the user.

A lot of projects is either using $HOME/.config or a dedicated folder in home:
E.g.
![screen shot 2015-05-09 at 11 35 00 pm](https://cloud.githubusercontent.com/assets/5354184/7550845/16c19746-f6a4-11e4-975c-774a62eb44bd.png)

lastquestion:
`$HOME/.fpp` is preferable to me... that way `fpp` doesn't have to deal with user issues.
proposed PR: https://github.com/facebook/PathPicker/pull/77

pcottle:
[...]

PR #77, opened 2015-05-10
use ~/.fpp instead of cluttering up ~

lastquestion:
:wrench:

### DIFF
--- a/src/choose.py
+++ b/src/choose.py
@@ -17,8 +17,8 @@ import output
 import screenControl
 import logger
 
-PICKLE_FILE = '~/.fbPager.pickle'
-SELECTION_PICKLE = '~/.fbPager.selection.pickle'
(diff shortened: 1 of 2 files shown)


lastquestion:
fixes https://github.com/facebook/PathPicker/issues/17

pcottle:
Sweet! I'll resolve the merge conflict and rebase this today, but yeah I think this is the best option compared to using `/tmp` and dealing with multiple user issues.

I might make another module here though that both can import so we dont duplicate these constants anywhere (which was a bad idea to begin with)

pcottle:
Alright we contain everything in `~/.fpp/` now, much cleaner. thanks @lastquestion for the suggested fix and pull request! :tada:

lastquestion:
I think the import module is a good idea but I didn't want to make a big PR. :+1:

PR #80, opened 2015-05-10
Added support for custom regexes

saevarb:
I saw this tool and thought it would fit my workflow really well -- if it was just a bit more pliable. 

So I added support for custom regular expressions. It's kind of .. bolted on, which is why I'd appreciate any pointers on how to add this functionality so it 'fits better', so to speak. That's assuming you are interested in merging this at all, however. 

There are tools [percol](https://github.com/mooz/percol) which let you do interactive grepping, but they don't let you filter out the results manually. Adding this functionality basically gives you a tool that works somewhere in between grep and percol, but with the ability to manually filter out what you want.

An example usage scenario would be something like:

    $ pacman -Qdt | ./fpp --regex "(^[\w-]*)"

`pacman` is the package manager for Arch Linux, and this lets me list orphaned packages and then manually select those I want to remove, hit `c` and then type `pacman -R` and they are gone.

Maybe this isn't an intended use-case for this tool, and if so, that's obviously fine. I may just keep my fork around anyway.

### DIFF
--- a/fpp
+++ b/fpp
@@ -23,6 +23,8 @@ BASEDIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 PYTHONCMD="python"
 # we need to handle the --help option outside the python
 # flow since otherwise we will move into input selection...
+# Hack to be able to access argument following '-r'/'--regex'
+i=0
 for opt in "$@"; do
   if [ "$opt" == "--debug" ]; then
     echo "Executing from '$BASEDIR'"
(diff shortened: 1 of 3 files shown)


pcottle:
> pacman is the package manager for Arch Linux, and this lets me list orphaned packages and then manually select those I want to remove, hit c and then type pacman -R and they are gone.

Ahh interesting, yeah I hadn't thought to use this for things that aren't paths / files! However something like:
(Pick)[https://robots.thoughtbot.com/announcing-pick]
(Bash Dialog)[https://github.com/ramses0/bash-dialog]
or
(fzf)[https://github.com/junegunn/fzf]

Might work better for arbitrary matches, since they have fuzzy-matching and all sorts of other goodies. We are going to be adding on file validation soon (where we actually do an `os.path.isfile`) as well, so this fork might become harder to maintain as we narrow in on the file case.

Anyways interesting fork though! Thanks for your interest in the project :smile:

PR #86, opened 2015-05-12
Test with Travis CI

metavida:
This fixes #82 

[Build Status](https://travis-ci.org/metavida/PathPicker)

I opted to test a broad range of python versions (2.6 through 3.4) even though the README states that PathPicker "requires Python >2.6 and <3.0", Because the test coverage currently passes with all of those versions enabled I opted to test them all, for now. I have no vested interest in testing all of these versions, so no hard feelings if you want to only test 1 version before merging this pull.

PR #87, opened 2015-05-12
Added tox for testing multiple version of python

### LINKED ISSUE #81: Testing multiple version of python.
There's an excellent tox package that can test multiple version of python in one command.
The only requirement is for the user to have the version installed.

Say you have all of this:
```
python26 -V
python27 -V
python33 -V
python34 -V
```

You can run one command to automatically test all of it:
```
tox
```

I already have it running in my local branch, let me know if you think this is good and I'll create a PR.

zkanda:
Fixes #81


pcottle:
Do you mind if we chose #86 over this? My only hesitation with tox is that then we introduce a dependency and need setup.py, `pip`, etc etc. I kind of like how this project is a leaf node right now in the dependency graph, and since Travis CI provides the same functionality I think it's a lighter-weight solution

Been loving your other PR's though @zkanda, lemme know if you disagree

PR #89, opened 2015-05-12
Fix syntax error in dictionary initialization when using python 2.6

### LINKED ISSUE #88: python 2.6 syntax error
OS: Centos 6
Python version: 2.6.6
```
getting input
  File "/home/hegazy/github/PathPicker/src/choose.py", line 16, in <module>
    import screenControl
  File "/home/hegazy/github/PathPicker/src/screenControl.py", line 29
    mapping = {i: chr(i) for i in range(256)}
                           ^
SyntaxError: invalid syntax
```
This is the trace that I got and there was a weird behavior that one file just opened in my text editor without seeing any list.

when I change the `PYTHONCMD="python"` to use python 2.7.6 `PYTHONCMD="python2.7"`It works just fine.

hegazy:
Resolves #88 by using the old way of initializing dictionaries
and abandon the new syntax because the syntax used is only available
in python2.7 onwards as stated here
https://www.python.org/dev/peps/pep-0274/

### DIFF
--- a/src/screenControl.py
+++ b/src/screenControl.py
@@ -26,7 +26,7 @@ import logger
 CHROME_MIN_X = 5
 CHROME_MIN_Y = 0
 
-mapping = {i: chr(i) for i in range(256)}
+mapping = dict((i, chr(i)) for i in range(256))
 mapping.update((value, name[4:]) for name, value in vars(curses).items()
                if name.startswith('KEY_'))
 # special exceptions


pcottle:
thanks a bunch @hegazy ! Yeah once we get multiple versions of python tested (and more test coverage on the UI bits) hopefully we can avoid these in the future

hegazy:
Shit, the commit is not assigned to my profile -_-

pcottle:
I think there might have been a merge conflict which means I had to manually resolve it. do you know if theres a way to do manual merges but attribute the author to someone else? id be curious since definitely want to give you credit

hegazy:
Actually, the problem is from my side. Git wasn't configured correctly at my machine like discussed here https://help.github.com/articles/why-are-my-contributions-not-showing-up-on-my-profile/#common-reasons-that-contributions-are-not-counted
Can I revert it with a another commit? :-)

pcottle:
Uh sure, if you want to open a PR that reverts and then reverts the revert, go for it

hegazy:
Sorry, that would be very silly to do, I like the clean history. 
Anyway, Thank you very much for the great project.

pcottle:

PR #90, opened 2015-05-12
Unicode support

zkanda:
Don't merge this yet, still needs more work.

But surprise everyone:
Unicode seems to be impossible to support in Python 2 curses: See this bug http://bugs.python.org/issue18118 :hankey: 

I temporarily selected python3 as the interpreter to make it possible to work on it.

cc @mknecht

### DIFF
--- a/src/test.py
+++ b/src/test.py
@@ -192,6 +192,22 @@ fileTestCases = [{
     'match': True,
     'num': 0,
     'file': '~/src/categories/NSDate+Category.h'
+}, {
+    # Unicode
+    'input': u'\xe4.py',
+    'match': True,
+    'num': 0,
+    'file': u'\xe4.py',
+}, {
+    'input': u'test.py\xe4',
+    'match': True,
+    'num': 0,
+    'file': u'test.py\xe4',
+}, {
+    'input': u'test\xe4.\xe42',
+    'match': False,
+    'num': 0,
+    'file': u'test\xe4.\xe42',
 }]
 
 prependDirTestCases = [
--- a/fpp
+++ b/fpp
@@ -20,7 +20,7 @@ while [ -h "$SOURCE" ]; do
 done
 BASEDIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 
-PYTHONCMD="python"
+PYTHONCMD="python3"
 
 function doProgram {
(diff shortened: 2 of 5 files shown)


zkanda:
Fixes #90

src/test.py
```diff
@@ -192,6 +192,22 @@
...
+}, {
+    'input': u'test.py\xe4',
+    'match': True,
+    'num': 0,
+    'file': u'test.py\xe4',
+}, {
+    'input': u'test\xe4.\xe42',
+    'match': False,
+    'num': 0,
+    'file': u'test\xe4.\xe42',
```
pcottle: :+1: pretty hardcore to have unicode file extensions haha

pcottle:
Hrm, I'm wondering if unicode support is worth dropping python2 support, or at least making it conditional on the version of python perhaps.

we'd also have to handle unix distributions where I believe the default `python` is python3 and only `python2` exists (not python3 command, unlike installing with homebrew on mac).

Glad you got this working though @zkanda, how does the screen output look like? Also is the output to the shell script correct as well?

lastquestion:
we can totally detect unicode even in python2, so a possible approach that would be pretty solid IMHO is to detect, quit and print out "you need python3 yo: here is #{python --version}"

for ppl with unicode, they'll know that to get what they really want they'll have to get python3, for people who couldn't care less, fpp Will Just Work for almost any unix-y environment, and for people who we detected wrong, they at least know how to fix it or google the right term to fix. (or we can put some info in the diagnostic, too)

pcottle:
Yeah good point @lastquestion . We already have a big error warning where we see git abbreviated paths:

screen shot 2015-05-13 at 11 10 50 am

so maybe we could do something similar for unicode

src/test.py
akroxaalyaa: S

pcottle:
Closing this out @zkanda since its overdue for a rebase (and we can track the feature in the task), but now that we are python 3.0 compatible i think it'd be easy-er to add this and just throw up a warning for utf8 files on python <3.0!

at least handling the display of utf8 would get us headed in the right direction

(also sorry to aggro-close but since Facebook internally tracks the average age of open pull requests and this is one of the only open PR's still, its driving our average up a lot :P )

PR #92, opened 2015-05-13
fix printing off the screen exceptions

lastquestion:
`maxy` come from `screenControl.getChromeBoundaries` which calls `stdscr.getmaxyx`, which returns the size in 0 indexed values. if the height of the screen is `10`, then valid `y`s are `0-9`. So you cannot print if `y == maxy`.

Secondly, sometimes exceptions will be thrown if curses prints the cursor off the screen, which occurs when `addstr` reads `\n` on the very last line of text. We already know what line we're printing and we control the x,y, so let's just strip that off during input processing.

### DIFF
--- a/src/processInput.py
+++ b/src/processInput.py
@@ -24,6 +24,10 @@ def getLineObjs():
     lineObjs = {}
     for index, line in enumerate(inputLines):
         line = line.replace('\t', '    ')
+        #remove the new line as we place the cursor ourselves for each
+        #line. this avoids curses errors when we newline past the end of the
+        #screen
+        line = line.replace('\n', '')
         formattedLine = FormattedText(line)
         result = parse.matchLine(str(formattedLine))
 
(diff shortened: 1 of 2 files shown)


src/processInput.py
```diff
@@ -24,6 +24,10 @@ def getLineObjs():
     lineObjs = {}
     for index, line in enumerate(inputLines):
         line = line.replace('\t', '    ')
+        #remove the new line as we place the cursor ourselves for each
+        #line. this avoids curses errors when we newline past the end of the
```
pcottle: this will break autopep8 but i can fix in a followup

pcottle:
> when addstr reads \n on the very last line of text. 

ahh hadn't thought about that case -- thanks! all about those off-by-one errors...

src/processInput.py
lastquestion: hm, I ran ./buildandtest I think. Sorry. next time I will make sure to run it

PR #96, opened 2015-05-13
fix symlink in README

silenc3r:
ln expects absolute paths (at least on linux)


pcottle:
Yeah this works for me! We can handle either relative or absolute symlinks, but if linux needs absolute ones then I guess we can go for it

PR #97, opened 2015-05-14
Example commit showing adding test coverage

pcottle:
Hey guys! I wanted to show a demo of the new test coverage I added this afternoon.

It unfortunately requires a fair amount of dependency injection (since we have to wrap both the curses module and the actual standard screen object) but we have a pretty sweet workflow setup.

Basically you can add a plaintext file of input that we will fake-pipe into the program, and then a series of keypresses. We'll record all the characters hitting the screen, compose a list of the screens the user saw, and then allow you to play those back and compare the expected output.

It doesn't support color or underlining yet (probably should) but now at least we have a lot more test coverage on our UI code. And this will be automatically tested with the Travis CI added this afternoon too!

### DIFF
--- a/src/__tests__/expected/selectTwoCommandMode.txt
+++ b/src/__tests__/expected/selectTwoCommandMode.txt
@@ -9,8 +9,8 @@
 ==========================================================
 Files you have selected:
 ==========================================================
-./README.md
-/Users/pcottle/Dropbox (Facebook)/wip/PathPicker/src/__tests__/__init__.py
+/foo/bar/README.md
+/foo/bar/fpp
 ==========================================================
 Type a command below! Files will be appended or replace $F
 Enter a blank line to go back to the selection process
--- a/src/__tests__/expected/selectTwoCommandMode.txt
+++ b/src/__tests__/expected/selectTwoCommandMode.txt
@@ -1,0 +1,30 @@
+
+
+
+
+
+
+
+
+==========================================================
+Files you have selected:
+==========================================================
+./README.md
+/Users/pcottle/Dropbox (Facebook)/wip/PathPicker/src/__tests__/__init__.py
+==========================================================
+Type a command below! Files will be appended or replace $F
+Enter a blank line to go back to the selection process
+==========================================================
+..........................................................
+
+
+
+
+
+
+
+
+
+
+________________________________________________________________________________
+command examples: | git add | git checkout HEAD~1 -- | mv $F ../here/ |
--- a/src/__tests__/inputs/absoluteGitDiff.txt
+++ b/src/__tests__/inputs/absoluteGitDiff.txt
@@ -1,0 +1,15 @@
+ /foo/bar/README.md                      |  8 ++++-
+ /foo/bar/fpp                            |  6 ++--
+ /foo/bar/src/__tests__/__init__.py      |  0
+ /foo/bar/src/__tests__/cursesForTest.py | 45 ++++++++++++++++++++++++++++
+ /foo/bar/src/__tests__/initTest.py      | 28 ++++++++++++++++++
+ /foo/bar/src/__tests__/screenForTest.py | 67 ++++++++++++++++++++++++++++++++++++++++++
+ /foo/bar/src/charCodeMapping.py         | 20 +++++++++++++
+ /foo/bar/src/choose.py                  | 15 ++++++++--
+ /foo/bar/src/colorPrinter.py            | 21 ++++++++-----
+ /foo/bar/src/cursesAPI.py               | 40 +++++++++++++++++++++++++
+ /foo/bar/src/format.py                  |  4 +--
+ /foo/bar/src/processInput.py            |  7 +++++
+ /foo/bar/src/screenControl.py           | 28 +++++++-----------
+ /foo/bar/src/screenFlags.py             | 34 +++++++++++++++++++++
+ 14 files changed, 290 insertions(+), 33 deletions(-)
--- a/src/__tests__/testScreen.py
+++ b/src/__tests__/testScreen.py
@@ -32,6 +32,10 @@ screenTestCases = [{
 }, {
     'name': 'selectDownSelectInverse',
     'inputs': ['f', 'j', 'f', 'A'],
+}, {
+    'name': 'selectTwoCommandMode',
+    'inputs': ['f', 'j', 'f', 'c'],
+    'pastScreen': 1
 }]
 
 
--- a/src/__tests__/testScreen.py
+++ b/src/__tests__/testScreen.py
@@ -34,6 +34,7 @@ screenTestCases = [{
     'inputs': ['f', 'j', 'f', 'A'],
 }, {
     'name': 'selectTwoCommandMode',
+    'input': 'absoluteGitDiff.txt',
     'inputs': ['f', 'j', 'f', 'c'],
     'pastScreen': 1
 }]


src/__tests__/expected/selectTwoCommandMode.txt
```diff
@@ -0,0 +1,30 @@
...
+
+
+
+
+
+==========================================================
+Files you have selected:
+==========================================================
+./README.md
+/Users/pcottle/Dropbox (Facebook)/wip/PathPicker/src/__tests__/__init__.py
```
pcottle: oh wait, this wont work outside of my directory because of the relative imports. let me update...

src/__tests__/testScreen.py
```diff
@@ -32,6 +32,10 @@
 }, {
     'name': 'selectDownSelectInverse',
     'inputs': ['f', 'j', 'f', 'A'],
+}, {
+    'name': 'selectTwoCommandMode',
+    'inputs': ['f', 'j', 'f', 'c'],
+    'pastScreen': 1
```
pcottle: here we denote we want the second screen we saw -- aka the screen we saw once we entered command mode

src/__tests__/expected/selectTwoCommandMode.txt
```diff
@@ -0,0 +1,30 @@
...
+
+
+
+
+
+
+==========================================================
+Files you have selected:
+==========================================================
+/foo/bar/README.md
```
pcottle: alright here we go. see! this is what a user would see as if they were in the UI

src/__tests__/inputs/absoluteGitDiff.txt
```diff
@@ -0,0 +1,15 @@
+ /foo/bar/README.md                      |  8 ++++-
```
pcottle: and the input

src/__tests__/testScreen.py
```diff
@@ -32,6 +32,11 @@
 }, {
     'name': 'selectDownSelectInverse',
     'inputs': ['f', 'j', 'f', 'A'],
+}, {
+    'name': 'selectTwoCommandMode',
+    'input': 'absoluteGitDiff.txt',
+    'inputs': ['f', 'j', 'f', 'c'],
```
pcottle: and the key sequence we perform

pcottle:
cc @metavida @lastquestion @frantic  @zkanda ^^

Also travis reporting back to me that the relative path resolution failed the tests was pretty sweet, since they passed locally!

src/__tests__/inputs/absoluteGitDiff.txt
```diff
@@ -0,0 +1,15 @@
...
+ /foo/bar/src/__tests__/screenForTest.py | 67 ++++++++++++++++++++++++++++++++++++++++++
+ /foo/bar/src/charCodeMapping.py         | 20 +++++++++++++
+ /foo/bar/src/choose.py                  | 15 ++++++++--
+ /foo/bar/src/colorPrinter.py            | 21 ++++++++-----
+ /foo/bar/src/cursesAPI.py               | 40 +++++++++++++++++++++++++
+ /foo/bar/src/format.py                  |  4 +--
+ /foo/bar/src/processInput.py            |  7 +++++
+ /foo/bar/src/screenControl.py           | 28 +++++++-----------
+ /foo/bar/src/screenFlags.py             | 34 +++++++++++++++++++++
+ 14 files changed, 290 insertions(+), 33 deletions(-)
```
pcottle: this is the input file

metavida:
Nice! CI for the win!

Love the ability to pass in an arbitrary full text input. Having to provide the full expected output is a bit frail (e.g. any change at all to the real output will break all existing tests) but that's definitely something that can be improved on & having better test coverage is a big win in the short term.

pcottle:
Yeah I'm hoping I can build some kind of interactive mode where you can approve changes to the expected output. We have a bunch of codegen tests like that at FB, and its prettylightweight when its just a diff preview and you tap y/n

Merging this in!

lastquestion:
A couple of thoughts. This is a really solid mocking approach, IMHO.
it would have been nice to get introspection on the code that runs this; it's pretty hard to read the diffs per commit and there isn't a good branch point to do a complete diff. I hadn't yet merged upstream with my master, so I was able to read the diffs clearly, but that's not always true for everybody.

1. the curses mock is really needed and good.

2. Full integration testing - which I would claim this to be - is super important. :clap: :clap: I tend to feel that it's difficult to really cover and exercise all corner cases via integration testing, but on the flip side, just because you test all the components doesn't mean it all works. LOL. So it's super important.

It's an interesting conversation, and I generally believe that unit testing is more efficient, tends to be less fragile, and provides more value. I'd like for us to write some unit tests utilizing the new curser mock (which is excellent) and the input module (which is again great).

To give a specific example, I would be very interested in writing tests that would exercise the chrome, the scroll bar, the command entry mode, the color printing, the forthcoming x-mode, etc. in a unit testing fashion. 

For chrome, for example, I would initialize the chrome object, send it the curses wrapper, and then assert that it called addstr at the right places and with the right strings. Then when I say move the scrollbar I assert it clears the right things and then it draws the new scroll bar at the right place. 

That's not to say the scroll bar is not testable in a replay system, it's just more brittle.

3. I tend to believe in code versus configuration. It's a bit of a `gulp` versus `grunt` conversation. For example, I'd probably want to write tests in python that each specify the inputs and expected outputs, as opposed to configuration based auto-compare against certain files in a directory.

pcottle:
> it would have been nice to get introspection on the code that runs this; it's pretty hard to read the diffs per commit and there isn't a good branch point to do a complete diff. I hadn't yet merged upstream with my master, so I was able to read the diffs clearly, but that's not always true for everybody.

Yeah sorry, I always get freaked out about losing work so I push often. I had a whole separate branch with the ~10 diffs needed for this stacked, but I guess since I rebased instead of squashing+merging it was harder to read. I didn't realize people read commits that closely! I'll try to do better in the future

> I tend to feel that it's difficult to really cover and exercise all corner cases via integration testing, but on the flip side, just because you test all the components doesn't mean it all works. LOL. So it's super important.

Yeah we've actually seen a big swing towards integration testing internally at FB, and I'm a huge fan of it. Some unit tests (especially those with automocking) only cover a very very shallow layer of code functionality, and the gaps between code (which are the hardest to unit test) is where the majority of issues can be found. But this way we get piped stdin, keyboard input, and **frame-by-frame** inspection of the output which is pretty nice

> For chrome, for example, I would initialize the chrome object, send it the curses wrapper, and then assert that it called addstr at the right places and with the right strings. Then when I say move the scrollbar I assert it clears the right things and then it draws the new scroll bar at the right place.

Yeah the curses / screen wrapper totally allows us to write those tests as well -- so I think it's good either way. My only worry about asserting individual `addstr` calls is that you can achieve the final output of the screen in multiple ways (with many different orderings or types of addstr calls), so its fairly implementation detaily. Hence why we only compare the final screen, regardless of *how* it was achieved

I think tomorrow I'll try to add a mode that records some kind of attribute-based logic, so it shows which parts are underlined and which parts arent.

then filesystem validation and that command-passing mode if appropriate

lastquestion:
> Yeah sorry, I always get freaked out about losing work so I push often. I had a whole separate branch with the ~10 diffs needed for this stacked, but I guess since I rebased instead of squashing+merging it was harder to read. I didn't realize people read commits that closely! I'll try to do better in the future

I think you had a branch but you merged it and I didn't know which branch it was when I first started reading. I was hoping to read a github branch:branch diff, and I ended up reading my master against upstream master which was clear and not per commit. I think we all commit pretty often nowadays which is why the value of the branch diff is so high now. Like I think `screentestDI` branch was basically what I was looking for but I couldn't find a merge commit for it in the first couple of commits in master so I got lost. 

> Yeah we've actually seen a big swing towards integration testing internally at FB, and I'm a huge fan of it. 
I haven't decided either way how I feel about it yet. This is getting off topic and I would be totally happy to get a beer and discuss this off-line. I do feel that there is value in full stack integration tests but that we don't yet as an industry have good knowledge about how to write unit-integration tests. I want to test the peanut butter glue that lives between the two slices of toast in isolation, as opposed to testing the entire PBJ sandwich as a whole.

> My only worry about asserting individual addstr calls is that you can achieve the final output of the screen in multiple ways (with many different orderings or types of addstr calls), so its fairly implementation detaily

Right, right, the tests would still wait for the entire result to be composed, just like the current integration test does, but would only assert on the correct values of parts of the screen that is "interesting to that test", as opposed to the complete screen. 

> I think tomorrow I'll try to add a mode that records some kind of attribute-based logic, so it shows which parts are underlined and which parts arent.
For this I would suggest actually writing out the format specifiers using the code in `colorprinter`; e.g. when you see a setattr in the curses mock, write out the corresponding format specifier. You only need to write a special case when addstr directly holds an attribute too, and that can be easily achieved by saving and then restoring the existing attribute. This way, the resulting recoded output can actually just be catt'ed on the console and get the real, colored and underlined text you're supposed to see. additionally your input can contain the attribute format specifiers as well.

PR #99, opened 2015-05-14
add cmd support (-c, --command)

lastquestion:
I want to add a test for this, but because it spans the whole batch script -> python thing, I want some guidance.

Passes pep8.

### DIFF
--- a/fpp
+++ b/fpp
@@ -21,20 +21,22 @@ done
 BASEDIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 
 PYTHONCMD="python"
+OPTS="$@"
 
 function doProgram {
   # process input from pipe and store as pickled file
-  $PYTHONCMD "$BASEDIR/src/processInput.py" "$@"
+  $PYTHONCMD "$BASEDIR/src/processInput.py" $OPTS
   # now close stdin and choose input...
   exec 0<&-
-  $PYTHONCMD "$BASEDIR/src/choose.py" "$@" < /dev/tty
+
+  $PYTHONCMD "$BASEDIR/src/choose.py" $OPTS < /dev/tty
   # execute the output bash script
   sh ~/.fpp/.fpp.sh < /dev/tty
 }
 
 # we need to handle the --help option outside the python
 # flow since otherwise we will move into input selection...
-for opt in "$@"; do
+for opt in OPTS; do
   if [ "$opt" == "--debug" ]; then
     echo "Executing from '$BASEDIR'"
   elif [ "$opt" == "--python3" ]; then
--- a/src/__tests__/screenTestRunner.py
+++ b/src/__tests__/screenTestRunner.py
@@ -10,10 +10,13 @@ from __future__ import print_function
 import curses
 import sys
 import os
+
 sys.path.insert(0, '../')
 
 import choose
 import processInput
+from screenFlags import ScreenFlags
+
 from screenForTest import ScreenForTest
 from cursesForTest import CursesForTest
 
@@ -29,19 +32,21 @@ def getLineObjsFromFile(inputFile):
 
 
 def getRowsFromScreenRun(
-      inputFile,
-      charInputs,
-      screenConfig={},
-      printScreen=True,
-      pastScreen=None):
+        inputFile,
+        charInputs,
+        screenConfig={},
+        printScreen=True,
+        pastScreen=None):
     lineObjs = getLineObjsFromFile(inputFile)
     screen = ScreenForTest(
         charInputs,
         maxX=screenConfig.get('maxX', 80),
         maxY=screenConfig.get('maxY', 30),
     )
+    # mock our flags with an empty command line for now
+    flags = ScreenFlags.initFromArgs([])
     try:
-        choose.doProgram(screen, CursesForTest(), lineObjs)
+        choose.doProgram(screen, flags, CursesForTest(), lineObjs)
     except StopIteration:
         pass
     if printScreen:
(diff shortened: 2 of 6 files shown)


lastquestion:
One thing: in `fpp`: I'd like to remove the matching of `record` and do that in `choose.py`, but by doing so, we'd slightly change behavior: we wouldn't print the `record` text until *after* `processInput` finishes. @pcottle: are you fine with that? IMHO it would clean up the code for longer term. Or, we can instantiate a copy of `ScreenFlags` in `processInput` and do the outputting there, if you'd like.

lastquestion:
Another thing: I wanted to support prefilling in the command in *command-mode*, but it was very difficult to do so. Basically, there isn't a default built in text-box in `ncurses`, and the only thing that exists that sort of does what we would want is `urwid` (https://github.com/wardi/urwid), but it's way too heavyweight and the text box there still doesn't support `insert-mode`.

I think we should code up our own text entry widget and turn back on the cursor for the command mode if we want to support prefilled in commands from the command line in that screen. I like the idea of it and we should do it in a separate, follow up PR.

src/__tests__/screenTestRunner.py
```diff
@@ -29,19 +32,21 @@ def getLineObjsFromFile(inputFile):
...
+        printScreen=True,
+        pastScreen=None):
     lineObjs = getLineObjsFromFile(inputFile)
     screen = ScreenForTest(
         charInputs,
         maxX=screenConfig.get('maxX', 80),
         maxY=screenConfig.get('maxY', 30),
     )
+    # mock our flags with an empty command line for now
+    flags = ScreenFlags.initFromArgs([])
```
pcottle: ah i didnt realize it was so easy to mock arg-based parsing... aka i didn't realize `argparse` could take in an empty array and not complain

src/choose.py
```diff
@@ -91,4 +91,8 @@ def setSelectionsFromPickle(selectionPath, lineObjs):
         output.writeToFile('echo ":D"')
         sys.exit(0)
     output.clearFile()
-    curses.wrapper(doProgram)
+    # we initialize our args *before* we curse up
+    # so we can benefit from the default argparse
+    # behavior:
+    flags = ScreenFlags.initFromArgs(sys.argv)
+    curses.wrapper(functools.partial(doProgram, flags=flags))
```
pcottle: you can also do:

```
lambda x:doProgram(x, flags)
```

do avoid the partial bind but nbd

lastquestion: that's cleaner, I was afraid because I didn't know what python level supported what.

lastquestion:
I'll rewrite to use the lambda and then it should be good to merge?

lastquestion:
also what's your thought about removing the match of `record` from `fpp` and moving it into `choose`

pcottle:
> I'll rewrite to use the lambda and then it should be good to merge?

yeah! i can do it as a followup if you want

> also what's your thought about removing the match of record from fpp and moving it into choose

sure thats fine, i kinda just had it there more for documentation anyways

pcottle:
actually ill just do the followups, want to get my hands on some code today anyways

pcottle:
Hrm @lastquestion i think the issue here is that bash includes the script dir as the first arg, and then python gets that by accident:

```
[pcottle:~/Dropbox (Facebook)/wip/PathPicker:master]$ ./fpp
the opt OPTS
Using old result...
i got args ['/Users/pcottle/Dropbox (Facebook)/wip/PathPicker/src/choose.py']
```

pcottle:
Alright I fixed that but still get errors when using other args:

[pcottle:~/Dropbox (Facebook)/wip/PathPicker:master]$ ./fpp --python3
Using old result...
usage: fpp [-h] [-r] [-ko] [-c COMMAND [COMMAND ...]]
fpp: error: unrecognized arguments: --python3

wonder if we can make it more permissive

pcottle:
think we had another bug:

```
     def initFromArgs(argv):
-        args = ScreenFlags.getArgParser().parse_known_args(argv)
+        (args, chars) = ScreenFlags.getArgParser().parse_known_args(argv)
         return ScreenFlags(args)
```

fpp
```diff
@@ -21,20 +21,22 @@ done
...
+
+  $PYTHONCMD "$BASEDIR/src/choose.py" $OPTS < /dev/tty
   # execute the output bash script
   sh ~/.fpp/.fpp.sh < /dev/tty
 }
 
 # we need to handle the --help option outside the python
 # flow since otherwise we will move into input selection...
-for opt in "$@"; do
+for opt in OPTS; do
```
pcottle: hrm i think this broken the argument parsing in bash, reverting

lastquestion:
Argh, yeah, revert this.

a) I forgot that we are not stripping the other arguments into `argparse`
b) there is something super weird with how "$@" is expanded in bash. 

Sorry, did you revert it already? to fix it, we  should add the other args to the argparse arguments so it doesn't blow up on --python3 etc.

for the other, I think the OPTS expands differently right? I didn't test with --python3 or --ko with -c at the same time for example. My bad again.

pcottle:
no need to revert -- i fixed everything in the commands above. i think everything works now! try fetching from upstream and let me know

lastquestion:
OK, cool. I'll fetch from upstream, double check. Sorry about that again. Is there a good way for us to test this? It's weird cuz it spans the batch to python land.

pcottle:
> Sorry about that again. Is there a good way for us to test this? It's weird cuz it spans the batch to python land.

i think ill add a UI test that just mocks the command being passed in and then checks we present the error message in the ui

pcottle:

Issue #98, opened 2015-05-14
Lingering text bug


xavierbeynon:
After unselecting an item, the trail of the selection text lingers
![demo](https://cloud.githubusercontent.com/assets/6413422/7626506/dbb7cf92-f9cf-11e4-8a1c-9e29116aef88.png)

lastquestion:
I noticed this earlier today. It was introduced by color support. I will open a PR for it today or tomorrow. Thanks for the bug report.

lastquestion:
Fixed in #100

pcottle:
How can you guys repro this? I actually can't repro at all, with python2 or python3.

lastquestion:
I need to fix up that PR.... 

Try to generate an input with one match at the end of the line. it should show up on a git diff for example.

pcottle:
![screen shot 2015-05-15 at 6 48 54 pm](https://cloud.githubusercontent.com/assets/1135007/7664238/0d7b4324-fb33-11e4-9049-fe18d6a0f481.png)

ah I see :P quite subtle!

pcottle:
Done! Thanks @lastquestion for the majority of the fix -- wanted to get this in ASAP before the weekend, since we just took out a bunch of Facebook ads for path picker and dont want any bugs if they install from master

lastquestion:
LOL did you really? I hope you aren't paying too much CPM. ^) Thanks, I'll look at your implementation now and follow up on everything that's happened, I saw your PRs. Thanks again for using PRs, it lets people follow along and understand the code evolution even if they can't get to stuff right that minute.

PR #100, opened 2015-05-14
fix bug where stale text wasn't being cleared

lastquestion:
I think master isn't pep8 compliant right now; both this PR and #99 needed to do changes to `screenTestRunner`, which is already in master.

Fixes the bug in #98. When a decorator is added to the line, the line length increases (obviously). When the decorated is removed, the line length shrinks, but because we don't clear the entire line, the old text remains. We fix this by printing blank, default background text, aka, 'erase' as many characters as needed to make the line match the "longest possible text line", which is the line + the longest decorator. (We only have one right now.)

We cache the blank text, so it's not expensive.

### DIFF
--- a/src/__tests__/testScreen.py
+++ b/src/__tests__/testScreen.py
@@ -77,16 +77,16 @@ class TestScreenLogic(unittest.TestCase):
         self.assertEqual(
             len(actualLines),
             len(expectedLines),
-            'Actual lines was %d but expected lines aws %d' % (
-                len(actualLines), len(expectedLines)),
+            'In file %s, actual lines was %d but expected lines was %d' % (
+                testName, len(actualLines), len(expectedLines)),
         )
         for index, expectedLine in enumerate(expectedLines):
             actualLine = actualLines[index]
             self.assertEqual(
                 expectedLine,
                 actualLine,
-                'Lines did not match:\n\nExpected:%s\nActual:%s' % (
-                    expectedLine, actualLine),
+                'Lines did not match in file %s:\n\nExpected:"%s"\nActual:"%s"' % (
+                    testName, expectedLine, actualLine),
             )
 
 if __name__ == '__main__':
--- a/src/format.py
+++ b/src/format.py
@@ -43,6 +43,7 @@ class SimpleLine(object):
 
 
 class LineMatch(object):
+    ARROW_DECORATOR = '|===>'
 
     def __init__(self, formattedLine, result, index):
         self.formattedLine = formattedLine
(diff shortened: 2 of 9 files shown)


lastquestion:
Also, I cleaned up some magic constants in `formattedText`.

src/__tests__/testScreen.py
```diff
@@ -77,16 +77,16 @@ def compareToExpected(self, testName, actualLines):
         self.assertEqual(
             len(actualLines),
             len(expectedLines),
-            'Actual lines was %d but expected lines aws %d' % (
-                len(actualLines), len(expectedLines)),
+            'In file %s, actual lines was %d but expected lines was %d' % (
+                testName, len(actualLines), len(expectedLines)),
```
pcottle: technically this is the test name and not the filepath, not sure if we want to make that clear

src/format.py
```diff
@@ -155,15 +156,30 @@ def updateDecoratedMatch(self):
...
+
+        # because decorators add length to the line, when the decorator
+        # is removed, we need to print blank space (aka "erase") the
+        # part of the line that is stale. calculate how much this is based
+        # on the max length decorator.
+        self.endingClearText = FormattedText(
+            FormattedText.getSequenceForAttributes(
+                FormattedText.DEFAULT_COLOR_FOREGROUND,
+                FormattedText.DEFAULT_COLOR_BACKGROUND, 0) +
+            " " * (self.getMaxDecoratorLength() - len(decoratorText)))
```
pcottle: hrm, im not a huge fan of this since then we are starting to print blank spaces after the lines **regardless* of whether we just become selected or not. you can see this in all the test input which has a bunch of trailing whitespace :-/

can we instead just dirty the line completely in screen control and use that to handle the refresh?

src/formattedText.py
```diff
@@ -42,8 +45,8 @@ def __str__(self):
     def parseFormatting(cls, formatting):
         """Parse ANSI formatting; the formatting passed in should be
         stripped of the control characters and ending character"""
-        fore = -1  # -1 default means "use default", not "use white/black"
-        back = -1
+        fore = cls.DEFAULT_COLOR_FOREGROUND
+        back = cls.DEFAULT_COLOR_BACKGROUND
```
pcottle: i do like these changes and a bunch of the others though!

pcottle:
Majority of this looks good, but let me try pulling this down and fixing this with simply dirtying the lines and redrawing completely rather than each line printing past its //actual// size

also keep in mind that printing trailing whitespace might lead to weird conditions -- say we printed the chrome and then printed a line next to the chrome **knowing** it fit inside the chrome. If it prints the trailing whitespace it might print over the chrome and lead to weird bugs

src/format.py
lastquestion: Yeah, that's a way better approach. I'll redo the PR.

lastquestion:
@pcottle : regarding the white space chrome issue - not an issue, the function `printUpTo` takes care of bounding no matter what kind of input is printed.

Dirtying the line is better though. I will redo the PR

src/__tests__/testScreen.py
lastquestion: I was kinda lazy. It would be cool if we can build the whole filepath. That way we can use fpp to pick it out. LOL. I'll change it to get the actual filepath.

pcottle:
> Dirtying the line is better though.

Yeah lemme know if that works -- should be pretty easy to tell our UI to just re-render the line that got changed

lastquestion:
Yah, I'll tackle this and the broken #99 later tonight.

pcottle:
Closing in favor of #109

PR #101, opened 2015-05-14
Use .values() when key is not needed.

dduong42:


pcottle:
:clap: :joy_cat: too easy! thanks for the PR

PR #104, opened 2015-05-14
Remove setController

dduong42:


pcottle:
yeah the set controller for the simple match was pretty silly -- i was clearly trying to maintain the interface even though theres no such thing in python

PR #105, opened 2015-05-15
Pythonic

dduong42:


pcottle:
im actually not a huge fan of the `@property` decorator :-/ it makes context switching super hard coming from php and js (what i work in all day) which doestnt have an equivalent concept

do you mind if we dont merge this one in? I think we should either be consistent and use them everywhere or just stick to the normal getters and setters

PR #106, opened 2015-05-15
Add command mode test

pcottle:
This is probably the best test coverage of command mode we can get for now. @lastquestion

### DIFF
--- a/src/__tests__/expected/selectCommandWithPassedCommand.txt
+++ b/src/__tests__/expected/selectCommandWithPassedCommand.txt
@@ -1,0 +1,30 @@
+
+
+
+
+
+
+
+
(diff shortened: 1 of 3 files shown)


lastquestion:
:clap: we should add a test for also adding commands that are picked up at the python level and make sure it eats the right commands and lets the other commands thru. I'll do that. Swamped today w/ real work tho.

pcottle:
> we should add a test for also adding commands that are picked up at the python level and make sure it eats the right commands and lets the other commands thru. I'll do that. Swamped today w/ real work tho.

yeah itd be nice to test that -- might be a bit tricky. no rush though, real work is the priority!

PR #108, opened 2015-05-16
Record attributes

pcottle:
Yay now we have attribute testing! we use `B` for blue, `_` for underlined, and blank for no changes.

Then we merge all the lines together so you can kind of read line-by-line what the formatting should look like


pcottle:
cc @lastquestion @frantic

frantic:
very clever!

lastquestion:
Yes, very clever. And clearer then a markup based approach. Interesting... could be useful in other applications and context too. :+1:

PR #109, opened 2015-05-16
PR #100 Resolves #98 Lingering text bug

### LINKED ISSUE #98: Lingering text bug

pcottle:
@lastquestion I took about half of your #100 PR and fixed it up with my interpretation -- I now realize why you made yours the way you did -- since we do "shrink back" after being selected, the only way we can make sure we dont have lingering characters is if we output to our full length.

however I kept the new logic a bit more contained -- I basically only called the blank output function when we are coming back from a selection state so less test files had to be changed and its (probably) more performant

I left some of the good parts of your PR out though, like your changes in `formattedText.py`. Mind following up with those?

### DIFF
--- a/src/__tests__/expected/selectDownSelectInverse.txt
+++ b/src/__tests__/expected/selectDownSelectInverse.txt
@@ -1,6 +1,6 @@
- README.md                      |  8 ++++-
+ README.md                      |  8 ++++-     
  fpp                            |  6 ++--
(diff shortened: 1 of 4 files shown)


pcottle:
> Mind following up with those?

that way you can get credit for this!

PR #110, opened 2015-05-16
Test with color input

pcottle:
This way we wont regress with color input! thx again @lastquestion


lastquestion:
:+1:. interesting design choice! I would have probably done something different with like, markup style, but this is highly readable, which is very nice! Nice.

pcottle:
Yeah markup style would have been interesting! At first I thought that would be hard since you would have to post process all the attribute maps but then I realized since you have the addstr function you can just do it inline :O that might have been more readable I guess too... Hrm. Anyways guess this works for now!

PR #111, opened 2015-05-16
Bunch more testing

pcottle:
got some scrollbar in there


lastquestion:
LOL I love it the tests are the sources of fpp itself. #inception

pcottle:
Yeah haha it's pretty meta. Might be annoying if we want to rename files one day (since some tests actually check if the input is a file) but I'll worry about that later! Might as well be efficient and use source code as the test input

PR #113, opened 2015-05-16
Double quote array expansions to avoid re-splitting element.

jbnicolai:
See https://github.com/koalaman/shellcheck/wiki/SC2068


pcottle:
Ahhhh that shellcheck project is cool. Thanks a bunch! Didn't know about the globbing expansion

jbnicolai:
Yeah shellcheck is amazing! :)

Issue #83, opened 2015-05-11
Path matching with filesystem validation/checking


jeremywiebe:
On branch master
    Changes not staged for commit:
      (use "git add <file>..." to update what will be committed)
      (use "git checkout -- <file>..." to discard changes in working directory)
    
      modified:   Hello Cordova.xcodeproj/project.pbxproj
      modified:   Hello Cordova.xcodeproj/project.xcworkspace/xcuserdata/jeremy.xcuserdatad/UserInterfaceState.xcuserstate
      modified:   Hello Cordova/config.xml
      modified:   www/cordova_plugins.js
      modified:   ../../plugins/fetch.json
      modified:   ../../plugins/ios.json
    
    Untracked files:
      (use "git add <file>..." to include in what will be committed)
      Hello Cordova/Plugins/com.phonegap.plugins.barcodescanner/
      Hello Cordova/Resources/scannerOverlay.xib
      www/plugins/com.phonegap.plugins.barcodescanner/
      ../../plugins/com.phonegap.plugins.barcodescanner/
    
    no changes add to commit (use "git add" and/or "git commit -a")

See also attached screenshot.  

![screen shot 2015-05-11 at 11 54 15 am](https://cloud.githubusercontent.com/assets/77138/7572645/a1f1698c-f7d4-11e4-819b-389985bab3c2.png)

pcottle:
Yeah, this is a pretty tough problem to solve since once you start splitting on whitespace, you'll end up matching on **everything**, especially when looking at code diffs where files and code are mixed together.

however I do think we could do something here behind an experimental flag that would basically be a super-matchy regex that actually checks on the filesystem level if the match is a file (and then only includes that).

so let me retitle this issue, since i think we'll move in that direction regardless

MissCrazyRarity:
It has issues as do we all and I need some help any sites known that I can go  to learn

jazen:
I moved my comment from #94 to this one to keep it all in one place.

I ran in my `~/git/dotfiles` repository and i'm using [thoughtbot's rcm](https://github.com/thoughtbot/rcm) tool for management which is why the files are not prefixed with a `.`:

```
On branch master
Your branch is up-to-date with 'origin/master'.
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

    modified:   irbrc
    modified:   pryrc
    modified:   rspec

[...]

PR #114, opened 2015-05-16
Filesystem validation

pcottle:
This is a first step towards addressing #83 -- we throw filesystem validation behind a flag on `matchLine` and (if its enabled), check that the file exists before returning the result.

This eliminates a bunch of false positives like the `and/or` in git status:
screen shot 2015-05-16 at 4 10 22 pm

And will open the door for us resolving the annoying relative paths in git status (versus top-level dirs).

A few open questions:
  * How should we turn this off for the screen testing?
  * Should this be behind a user flag or enabled for all? or perhaps able to be turned off?
  * for the git status issue, we actually need to change our prepend-path depending on a filesystem result -- so that might require some ugly threading

### DIFF
--- a/src/__tests__/expected/gitDiffWithValidation.txt
+++ b/src/__tests__/expected/gitDiffWithValidation.txt
@@ -1,0 +1,60 @@
+src/charCodeMapping.py         | 20 +++++++++++++
+BBBBBBBBBBBBBBBBBBBBBB                                                          
+src/choose.py                  | 15 ++++++++--
+_____________                                                                   
+src/colorPrinter.py            | 21 ++++++++-----
+___________________                                                             
+src/cursesAPI.py               | 40 +++++++++++++++++++++++++
+________________                                                                
+src/format.py                  |  4 +--
+_____________                                                                   
+src/processInput.py            |  7 +++++
+___________________                                                             
+/foo/bar/src/charCodeMapping.py         | 20 +++++++++++++
+                                                                                
+/foo/bar/src/choose.py                  | 15 ++++++++--
+                                                                                
+/foo/bar/src/colorPrinter.py            | 21 ++++++++-----
+                                                                                
+/foo/bar/src/cursesAPI.py               | 40 +++++++++++++++++++++++++
+                                                                                
+/foo/bar/src/format.py                  |  4 +--
+                                                                                
+/foo/bar/src/processInput.py            |  7 +++++
+                                                                                
+/foo/bar/src/screenControl.py           | 28 +++++++-----------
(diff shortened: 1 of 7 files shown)


pcottle:
alright, this is done now:

~~for the git status issue, we actually need to change our prepend-path depending on a filesystem result -- so that might require some ugly threading~~

so git status is handled, false positives handled, and all the testing is fixed. I punted on:
> Should this be behind a user flag or enabled for all? or perhaps able to be turned off?

I'm not superrrr happy with the implementation, since its a fair amount of bool param threading which I never like. However using dependency injection or some kind of global module seemed overkill, and adding a flag to disable it seemed backwards as well (it should be a straight improvement).

The one thing I wish we could resolve is `matchLine` does the `prependDir` check once, and then we have to do the prependDir check again (with the same bool param threaded in) to actually format the LineMatch object. Solving this would require a big refactor though...

lastquestion:
This feels good! Comments (mostly just agreeing with you verbosely):

1. You're right; the bool threading just doesn't feel good. I would advocate for a `settings` object that's globally accessible; we can repurpose the `ScreenFlags` class for this, because basically it's about the same kind of issue: "what are the configurations for this object". `ScreenFlags` currently figures out "what the user set as the configuration". We can combine the two, make it a singleton object, and call it a day. There is some complication about pickling: when loading a previous pickle, you'd also want to load the previous settings as well, but this can be resolved by pickling the settings as well.

2. Dealing with `matchline`: Yes, it's a bit annoying as well. I agree. A bigger refactor here would be needed, but it's also necessary. It's obviously clear because we do file stuff and parsing paths etc in a file called `format`, LOL. 

Both of these things are things I will try to make PRs for, but first I want to fix up the space printing dirty/non-dirty thing. :)

pcottle:
If we go to a global object that's accessible anywhere, we should also publish the writes globally (aka reimplement Flux but in Python).

A global object is way cleaner but it gets messy when mutating it -- anyone can write from anywhere (even in the middle of your function call) it's hard to track down.

I'd be down for ideas on cleaning this up though, since we have a ton of test coverage on it now

lastquestion:
I'd generally design a readonly singleton that can be assigned as a whole only at certain places; e.g. startup, "settings page exit", etc. whatever systems need to be restarted have a clean place to restart and you can rely on settings not changing under you at any time. 

But all this may be moot. it will be cleaner and less parameter-threading once formatter stops calling parser for file resolving. Let's take a look at it after I try to do that early this week. (Of course you are moving pretty fast, so I dunno whether you've had done a pass of refactoring by then...)

PR #116, opened 2015-05-17
Handle spaces in filenames

### LINKED ISSUE #115: FPP doesn't properly parse/handle paths with spaces in them
Related to #83, now that we have #114 lets add a fallback regex with spaces that only is on with filesystem inspection

pcottle:
Resolves #115

Need to handle the trimming (dont allow leading spaces?) and add test coverage, otherwise good

### DIFF
--- a/src/parse.py
+++ b/src/parse.py
@@ -41,13 +41,21 @@ FILE_NO_PERIODS = re.compile(''.join((
     '(\s|$|:)+'
 )))
 MASTER_REGEX_WITH_SPACES = re.compile(''.join((
-    # directories as normal
-    '(\/?([a-z.A-Z0-9\-_]+\/)+',
+    #begin the capture
+    '(',
+    #a leading / for absolute dirs if its there
+    '\/?',
+    # now we look at directories, where we allow either normal chars
+    # or a single whitespace char followed by legit chars
+    '(([a-z.A-Z0-9\-_]|\s[a-zA-Z0-9\-_])+\/)+',
     # you can have either a space followed by a real char OR real chars.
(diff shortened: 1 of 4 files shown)


src/parse.py
```diff
@@ -40,6 +40,25 @@
...
     '(\s|$|:)+'
 )))
+MASTER_REGEX_WITH_SPACES = re.compile(''.join((
+    #begin the capture
+    '(',
+    #a leading / for absolute dirs if its there
+    '\/?',
+    # now we look at directories, where we allow either normal chars
+    # or a single whitespace char followed by legit chars
+    '(([a-z.A-Z0-9\-_]|\s[a-zA-Z0-9\-_])+\/)+',
```
lastquestion: This is good, but unfortunately real directories can have multiple spaces. I'm not dinging the approach - the regex otherwise would overmatch on everything - and clearly it's an improvement. But '/just for you pcottle/' is a valid directory.  Something to think about.

pcottle: Oh sorry I should clarify -- the normal chars or single white space followed by the normal chars is the **inner match** of the directories, and we allow any number of inner matches for a given directory.

So you can have multiple spaces per folder name, you just can't end in a space or have two spaces in a row. It was a tricky hack since otherwise we would gobble up white space before any / and consider that a directory >___<

If you want you could send over a PR that adds testing or I can try to get to it this week!

lastquestion: Hey that's my fault, I read the comment but I didn't scroll up again and look at the tests as closely as I should have.

PR incoming to update the comment to explain that's inner matching.

PR #119, opened 2015-05-18
dirty lines instead of printing spaces

lastquestion:
This is way cleaner then the printing white space approach.

1. Use `erase` instead of `clear`. `erase` does not call `refresh`, whereas `clear` does. We need to draw before refreshing anyway, so we save on a `refresh` call which is (a) cleaner and (b) potentially flicker free on slow SSH connections, for ex.

2. We use delch to delete and 'reset' a line. It removes the character and the associated attribute at some index. I've also mocked this to mean exactly that in the test suite, so we don't need to have spaces anymore in the tests.

3. Rename `dirtyLines` to `dirtyAll` to match what it does: it dirties evrything and forces chrome to redraw.

4. Pep8 pass.

### DIFF
--- a/src/__tests__/expected/selectDownSelectInverse.txt
+++ b/src/__tests__/expected/selectDownSelectInverse.txt
@@ -1,6 +1,6 @@
- README.md                      |  8 ++++-     
+ README.md                      |  8 ++++-
  fpp                            |  6 ++--
- src/__tests__/__init__.py      |  0     
+ src/__tests__/__init__.py      |  0
  |===>src/__tests__/cursesForTest.py | 45 ++++++++++++++++++++++++++++
  |===>src/__tests__/initTest.py      | 28 ++++++++++++++++++
  |===>src/__tests__/screenForTest.py | 67 ++++++++++++++++++++++++++++++++++++++
--- a/src/format.py
+++ b/src/format.py
@@ -45,13 +45,16 @@ class LineMatch(object):
     ARROW_DECORATOR = '|===>'
 
     def __init__(self, formattedLine, result, index, validateFileExists=False):
+        self.controller = None
+
         self.formattedLine = formattedLine
         self.index = index
 
         (file, num, matches) = result
 
         self.originalFile = file
-        self.file = parse.prependDir(file, withFileInspection=validateFileExists)
+        self.file = parse.prependDir(file,
+                                     withFileInspection=validateFileExists)
         self.num = num
 
         line = str(self.formattedLine)
@@ -80,15 +83,10 @@ class LineMatch(object):
         # precalculate the pre, post, and match strings
         (self.beforeText, unused) = self.formattedLine.breakat(self.start)
         (unused, self.afterText) = self.formattedLine.breakat(self.end)
+
         self.updateDecoratedMatch()
-        self.controller = None
-        self.needsUnselectedPrint = False
 
     def toggleSelect(self):
-        if self.selected:
-            # we need to print ourselves blank at the end of the line
-            # to prevent the lingering text bug
-            self.needsUnselectedPrint = True
         self.setSelect(not self.selected)
 
     def setSelect(self, val):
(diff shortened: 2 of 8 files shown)


lastquestion:
I plan to open a follow on PR to use `pad` so we can actually paint the entire scroll buffer and cache it; `curses` takes care of painting onto the visible screen via `refresh`. It will greatly simplify the painting code and also be faster - we will never write a line twice until it changes highlight/hover/select state. I need to do some investigation on size limits though - like I'll try it with 10/50/100k lines and so, make sure it is performant. If so, it'll clean up a lot of code nicely.

src/format.py
```diff
@@ -45,13 +45,16 @@ class LineMatch(object):
...
+
         self.formattedLine = formattedLine
         self.index = index
 
         (file, num, matches) = result
 
         self.originalFile = file
-        self.file = parse.prependDir(file, withFileInspection=validateFileExists)
+        self.file = parse.prependDir(file,
+                                     withFileInspection=validateFileExists)
```
pcottle: why this change? was it over 80char?

src/screenControl.py
```diff
@@ -213,19 +213,20 @@ def __init__(self, flags, stdscr, lineObjs, cursesAPI):
         (self.oldmaxy, self.oldmaxx) = self.getScreenDimensions()
         self.mode = SELECT_MODE
 
-        self.simpleLines = []
```
pcottle: do we not use this for anything? :O

src/screenControl.py
```diff
@@ -490,27 +486,34 @@ def onEnter(self):
...
-        if self.linesDirty:
+        if self.dirty:
             self.printAll()
-        for index in self.dirtyIndexes:
-            self.lineMatches[index].output(self.colorPrinter)
-        if self.helperChrome.getIsSidebarMode():
-            # need to output since lines can override
-            # the sidebar stuff
-            self.printChrome()
+        else:
```
pcottle: nit: the above could just be an early return so we dont indent too much

src/screenControl.py
```diff
@@ -490,27 +486,34 @@ def onEnter(self):
...
+                y = miny + index + self.getScrollOffset()
+                if y >= miny or y < maxy:
+                    self.clearLine(y)
+                    self.lineObjs[index].output(self.colorPrinter)
+
+    def clearLine(self, y):
+        '''Clear a line of content, excluding the chrome'''
+        (minx, miny, maxx, maxy) = self.getChromeBoundaries()
+        for x in range(minx, maxx):
+            self.stdscr.delch(y, x)
```
pcottle: ahhhh cool!!!!

pcottle:
:tada: :tada: :tada: 
Woohoo! This is way better -- deleting that hacky whitespace at the end of the expected test output feels quite nice :D

I didn't know about delchar to be honest, clearly I should have read through more of the curses documentation. This is pretty sweet though, I like how we re-use the chrome boundaries to understand which lines to clear

Lemme merge this and followup with some nits

src/__tests__/expected/selectDownSelectInverse.txt
```diff
@@ -1,6 +1,6 @@
- README.md                      |  8 ++++-     
+ README.md                      |  8 ++++-
  fpp                            |  6 ++--
- src/__tests__/__init__.py      |  0
```
pcottle: :heart_eyes:

src/format.py
lastquestion: My version of pep8 complained and broke the line with `/`, which I didn't like so I used a new line instead. What version do you have installed?

src/screenControl.py
lastquestion: haha nah. I was surprised and double-grepped to be sure

src/screenControl.py
lastquestion: Yah. I'll change it around next PR

src/__tests__/expected/selectDownSelectInverse.txt
lastquestion: The best. :tada:

PR #120, opened 2015-05-18
better explanation of path space regex matching

lastquestion:


pcottle:
:tada: like the naming of "character class," I couldn't figure out a good thing to call that inner grouping. lemme add one or two more examples of what it does / does not match

Issue #118, opened 2015-05-18
Flag/configuration to shorten deep nested paths


QwertyManiac:
On limited screen width, or for heavily nested directory projects (such as apache/hadoop) the path picker output for a grep result becomes unusable as it only shows the directory paths than the filenames (the lines don't appear to wrap around).

It would be nice to show the paths in git style perhaps, with '…' replacing the initial directories so the shown focus may instead be on the filenames. This could also be configurable, if not desirable by all.

![screen shot 2015-05-18 at 10 27 36 am](https://cloud.githubusercontent.com/assets/31518/7674569/88227306-fd48-11e4-832d-66619319aba5.png)

lastquestion:
I thought about this and maybe we can support left and right arrow to page screen width to the left and right. I'd rather do that then '...' abbrev.

To do this, we'd probably want to use `pad` or implement a more generic dirty / scroll mechanism. Something I'll think about after trying to get a `pad` PR out. 

Not sure how @pcottle feels...

pcottle:
> I thought about this and maybe we can support left and right arrow to page screen width to the left and right. I'd rather do that then '...' abbrev.

Hrm, I'd rather actually do the abbreviation tbh so the relevant parts fit in the screen and i don't have to waste time scrolling over. it'd also be helpful if you had output like

```
./somethingsuperlongggggggggggggggggggggggggggggg/a1
./somethingsuperlongggggggggggggggggggggggggggggg/a2
./otherThingHere
./OtherThingThere
./somethingsuperlongggggggggggggggggggggggggggggg/a3
```

having to bounce back and forth between sides of the screen would be tough.

I don't think a flag would make sense here though -- we should just do this if your file match length is way too long compared to the chrome boundaries.

my only worry is how to implement the shortening logic -- @QwertyManiac would just showing the filename work for your case?

actually @lastquestion we could do something interesting here and more intelligent than git -- we could just use up all the available space, so itd be something like

```
./hadoop-mapreduce-|...|/client-hs/something-else.py | [f] is to select
./hadoop-mapreduce-|...|/client-hs/something-cool.py | [j] is to go down
```

aka write up all the way until the chrome

lastquestion:
Your argument is sound, it would be annoying to bounce back and forth.

[...]

PR #121, opened 2015-05-18
addstr asserts even with 0 to print, so break out early

lastquestion:
While investigating #118, I discovered a bug where we are causing `addstr` to throw a `curses` exception - at least on my mac osx, `addstr` excepts if you try to print outside the screen, even if the number of characters to print is 0.

also, added a test for super long file names and make sure they are truncated correctly.


src/__tests__/testScreen.py
```diff
@@ -173,7 +182,7 @@ def assertEqualLines(self, testName, actualLines, expectedLines):
             self.assertEqual(
                 expectedLine,
                 actualLine,
-                'Lines did not match for test %s:\n\nExpected:%s\nActual:%s' % (
+                'Lines did not match for test %s:\n\nExpected:%s\nActual  :%s' % (
```
lastquestion: add some spaces here to line up the expected and actual for easy visual diffing.

src/__tests__/inputs/longFileNames.txt
```diff
@@ -0,0 +1,2 @@
+longfilenamereallylongfilenameitssuperlongareyoutiredoftryingtoreadthisyet
```
pcottle: haha

src/__tests__/testScreen.py
```diff
@@ -83,6 +83,15 @@
     'input': 'gitDiffSomeExist.txt',
     'validateFileExists': True,
     'withAttributes': True,
+}, {
+    'name': 'longFileNames',
+    'input': 'longFileNames.txt',
+    'validateFileExists': False,
+    'withAttributes': False,
+    'screenConfig': {
```
pcottle: glad you found the screen config useful

src/__tests__/testScreen.py
pcottle: ah great idea -- hadnt thought about the character differences between the two labels. maybe we should use tabs...

pcottle:
I had a small merge conflict on the `Lines did not match for test %s:\n\nExpected:%s\nActual  :%` line from your other PR, so I just combined the two (visually lined them up and wrapped in quotes too)

lastquestion:
Yeah, that other fix would have caused a merge conflict. thanks for fixing it up! :+1:

PR #122, opened 2015-05-19
fix dirty lines properly, update tests for multi-screen recording

lastquestion:
While using FPP on a wide monitor, I discovered we were printing garbage using the just merged 'dirty' PR #119 

I then read the ORIGINAL documentation for `delch` from `man`:
http://dell9.ma.utexas.edu/cgi-bin/man-cgi?delch+3

notice it says 
```
all characters to the right  of  the  cursor  on the same line are moved to the left one
position and the last character on the line is  filled  with  a  blank.
```

This is not mentioned in the `ncurses` documentation for python, which merely says
```
window.delch([y, x])
Delete any character at (y, x).
```

:angry:

So, this PR:

* implements `clearLine` by writing " " using the default background color.
* monkey-patches `clearLine` during testing to actually do what it means, which is clear the output
* adds a wide-chrome test which tests selecting a match multiple times
* adds the capability to record every page to the test suite

### DIFF
--- a/src/__tests__/screenForTest.py
+++ b/src/__tests__/screenForTest.py
@@ -46,8 +46,9 @@ class ScreenForTest(object):
         return (self.maxY, self.maxX)
 
     def refresh(self):
-        # TODO -- nothing to do here?
-        pass
+        if self.containsContent(self.output):
+            # we have an old screen, so add it
+            self.pastScreens.append(dict(self.output))
 
     def containsContent(self, screen):
         for coord, pair in screen.items():
@@ -57,9 +58,6 @@ class ScreenForTest(object):
         return False
 
     def erase(self):
-        if self.containsContent(self.output):
-            # we have an old screen, so add it
-            self.pastScreens.append(self.output)
         self.output = {}
         for x in range(self.maxX):
             for y in range(self.maxY):
@@ -79,11 +77,6 @@ class ScreenForTest(object):
         for deltaX in range(len(string)):
             coord = (x + deltaX, y)
             self.output[coord] = (string[deltaX], self.currentAttribute)
-
-    def delch(self, y, x):
-        '''Delete a character. We implement this by removing the output,
-        NOT by printing a space'''
-        self.output[(x, y)] = ('', 1)
 
     def getch(self):
         return CHAR_TO_CODE[self.charInputs.pop(0)]
@@ -108,7 +101,16 @@ class ScreenForTest(object):
         return self.getRows(screen=self.pastScreens[pastScreen])
 
     def getRowsWithAttributesForPastScreen(self, pastScreen):
-        return self.getRowsWithAttributes(screen=self.pastScreens[pastScreen])
+        if pastScreen is None:
+            pastScreen = [self.getNumPastScreens()]
+        elif not isinstance(pastScreen, list):
+            pastScreen = [pastScreen]
+
+        pages = map(lambda screenIndex: self.getRowsWithAttributes(
+            screen=self.pastScreens[screenIndex]), pastScreen)
+        lines, attributes = zip(*pages)
+        return ([line for page in lines for line in page],
+                [line for page in attributes for line in page])
 
     def getRowsWithAttributes(self, screen=None):
         if not screen:
--- a/src/__tests__/testScreen.py
+++ b/src/__tests__/testScreen.py
@@ -36,7 +36,7 @@ screenTestCases = [{
     'name': 'selectTwoCommandMode',
     'input': 'absoluteGitDiff.txt',
     'inputs': ['f', 'j', 'f', 'c'],
-    'pastScreen': 1
+    'pastScreen': 3
 }, {
     'name': 'selectCommandWithPassedCommand',
     'input': 'absoluteGitDiff.txt',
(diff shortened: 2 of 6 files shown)


src/__tests__/screenForTest.py
```diff
@@ -108,7 +101,16 @@ def getRowsForPastScreen(self, pastScreen):
         return self.getRows(screen=self.pastScreens[pastScreen])
 
     def getRowsWithAttributesForPastScreen(self, pastScreen):
-        return self.getRowsWithAttributes(screen=self.pastScreens[pastScreen])
+        if pastScreen is None:
+            pastScreen = [self.getNumPastScreens()]
+        elif not isinstance(pastScreen, list):
+            pastScreen = [pastScreen]
+
+        pages = map(lambda screenIndex: self.getRowsWithAttributes(
```
pcottle: woof, whats with all this introduced complexity? why is pastScreen now a list? :*(

src/__tests__/testScreen.py
```diff
@@ -92,6 +92,17 @@
...
+    'name': 'dontWipeChrome',
+    'input': 'gitDiffColor.txt',
+    'withAttributes': True,
+    'validatesFileExists': False,
+    'inputs': ['DOWN', 'f', 'f', 'f', 'UP'],
+    'screenConfig': {
+        'maxX': 201,
+        'maxY': 40
+    },
+    'pastScreen': [0, 1, 2, 3, 4]
```
pcottle: oh I see. I'd rather have this be something like `pastScreens` rather than overloading the type of `pastScreen` and allowing lists as well as single screens.

that way the test checker can just become a loop of what we had earlier

src/__tests__/testScreen.py
```diff
@@ -196,4 +207,16 @@ def maybeMakeExpectedDir():
...
+    # this matches the intent of the function and makes sure we don't
+    # generate empty spaces in the output for comparing against the expected
+    # input
+
+    def clearLine(self, y):
+        (minx, miny, maxx, maxy) = self.getChromeBoundaries()
+        for x in range(minx, maxx):
+            self.stdscr.output[(x, y)] = ('', 1)
+
+    screenTestRunner.choose.screenControl.Controller.clearLine = clearLine
```
pcottle: :see_no_evil: this seems a bit hacky too. I know maybe its valid python but usually some kind of more obvious and more explicit dependency injection is better than overwriting the functionality of the module to be tested

pcottle:
Yep I definitely see the bug now, haha:
screen shot 2015-05-19 at 5 08 30 pm

this totally lame the `delch` documentation doesnt mention anything about that. some thoughts:

  * What if we implemented `clearLine` the same as before (with delchar) but went *right to left* and painted the chrome again afterwards?
  * If we switched back to the normal emptystring printing, how bad does the test output look like? I almost think its a bit more clear to do whitespace printing from the untoggled-input to the chrome boundaries than to do what we were doing earlier, which was painting basically an "arrow length" of whitespace after

pcottle:
Lets go with #123 over this if thats OK -- I think the monkeypatching for the tests would be ideal to avoid, and it'd be nice to use delchar in some capacity. I would be down to add the `pastScreens` option though!

src/__tests__/testScreen.py
lastquestion: I'm sort of OK with this specific instance of monkey patching, because I'd rather monkey patch `ncurses` itself. But you can't because it's a compiled module, so I had to replace the function in `screenController`.  If you'd like we can move this function into `cursesAPI`, and provide a wrapper around the existing `cursesAPI` when running against real `ncurses`. 

Actually this would all be better if the controller was separate from the view code here, but that's another refactor for another day.

src/__tests__/testScreen.py
lastquestion: Hm, I'm fine with that.

lastquestion:
Hmmm, on further thought, I think it's OK to use delchar because soon if we use `pad` we can completely separate the chrome output from the scroll output and there won't be any chrome to fuck up when you're erasing.

I would prefer option (a) here instead of (b)

pcottle:
By `pad` you mean curses windows and pads right? https://docs.python.org/2/howto/curses.html#windows-and-pads

Ah very cool! Man I really should have just spent an afternoon reading through the curses docs before rolling some of this infra on my own :O

i agree though, once the chrome is 100% its own pad then the delchar method will work flawlessly

pcottle:
By `pad` you mean curses windows and pads right? https://docs.python.org/2/howto/curses.html#windows-and-pads

Ah very cool! Man I really should have just spent an afternoon reading through the curses docs before rolling some of this infra on my own :O

i agree though, once the chrome is 100% its own pad then the delchar method will work flawlessly.

I guess the advantage of the pad is that we can reposition it off screen when its not displayed and move it back onscreen when appropriate? I guess our chrome is always shifting from the right to the bottom...

lastquestion:
Yah I mean pad... It also would let us paint the line match txt once and then scroll it up and down without having to reprint the words _i think_. I am going to experiment first and make sure the documentation isn't missing anything and that it'll work with large buffers (10k lines etc).

pcottle:
> it also would let us paint the line match txt once and then scroll it up and down without having to reprint the words i think. 

Ah interesting! I see what you mean -- yeah I wonder if its more efficient for us to hold our more minimal representation in memory or to just dump everything into the buffer and let them (more efficiently perhaps?) do the repainting when it moves.

lastquestion:
Unsure. It's why I'm gonna experiment. If the perf/memory is very close (as in ~5%) it's worth moving to pad because our code would be a lot simpler. But if the perf is a big difference then :(

PR #123, opened 2015-05-20
Followup to PR #119 Resolves #122 fix delchar bug

pcottle:
@lastquestion here is my minimal fix and the demo:

screen demo

### DIFF
--- a/src/screenControl.py
+++ b/src/screenControl.py
@@ -500,16 +500,28 @@ class Controller(object):
             self.printAll()
             return
         (minx, miny, maxx, maxy) = self.getChromeBoundaries()
+        didClearLine = False
         for index in self.dirtyIndexes:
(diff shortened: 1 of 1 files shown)


lastquestion:
Yeah, I think this is good, and also gives us a forward path for more stuff later where we can split the chrome with the scroll. 

I'll open a PR for the `pastScreens` and then we can add a test for wide like I had before. This will make sure we don't regress with the wide chrome and the `delchar`. God this dirty code has had 4 PRs against it already.

pcottle:
> I'll open a PR for the pastScreens and then we can add a test for wide like I had before. This will make sure we don't regress with the wide chrome and the delchar. 

ah good point, that'd be nice to add.

>  God this dirty code has had 4 PRs against it already.

Psh its not that bad!

PR #126, opened 2015-05-21
support multiple page recording for tests and add a wide chrome test

lastquestion:
Adding back the tests from the closed PR #122 
@pcottle, I created a new option `pastScreens`, which is close to `pastScreen`, but I think it's different enough; I'm not opinionated on the name so we can rename to something else too.

### DIFF
--- a/src/__tests__/testScreen.py
+++ b/src/__tests__/testScreen.py
@@ -36,7 +36,7 @@ screenTestCases = [{
     'name': 'selectTwoCommandMode',
     'input': 'absoluteGitDiff.txt',
     'inputs': ['f', 'j', 'f', 'c'],
-    'pastScreen': 1
+    'pastScreen': 3
 }, {
     'name': 'selectCommandWithPassedCommand',
     'input': 'absoluteGitDiff.txt',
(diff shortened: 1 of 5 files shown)


src/__tests__/testScreen.py
```diff
@@ -36,7 +36,7 @@
     'name': 'selectTwoCommandMode',
     'input': 'absoluteGitDiff.txt',
     'inputs': ['f', 'j', 'f', 'c'],
-    'pastScreen': 1
+    'pastScreen': 3
```
pcottle: is this change needed because of the change between `erase` and `refresh`?

src/__tests__/testScreen.py
```diff
@@ -92,6 +92,17 @@
...
+    'name': 'dontWipeChrome',
+    'input': 'gitDiffColor.txt',
+    'withAttributes': True,
+    'validatesFileExists': False,
+    'inputs': ['DOWN', 'f', 'f', 'f', 'UP'],
+    'screenConfig': {
+        'maxX': 201,
+        'maxY': 40
+    },
+    'pastScreens': [0, 1, 2, 3, 4]
```
pcottle: :heart_eyes:

pcottle:
:tada: :clap: wooo awesome! I do think its more explicit threading and more `.get(key, default)` values, but i think its cleaner from an internal API standpoint. thanks a bunch @lastquestion !

src/__tests__/testScreen.py
lastquestion: Yeah -  because we now take the screenshot when you "refresh". Before we only did it on erase and we don't always call erase on the entire screen only when we clear the entire screen. When we dirty a line (say from up down or j f) we just call refresh.

pcottle: Ah cool! good to know, yeah we can capture more intermediate states this way :+1:

PR #127, opened 2015-05-21
fix delchar again

lastquestion:
Yeah, I know.

Issue: the previous code deleted the characters (in reverse, so we wouldn't skip characters) but only up to the chrome. The problem is, delchar moves things to the left, so we end up with the chrome right next the cursor pos we started deleting at.

Added a GIGANTIC Q&A comment. 

To repro the original issue, make sure your terminal is >200 chars wide and try flipping a match, the chrome should jump right up against the end of the text line.

### DIFF
--- a/src/screenControl.py
+++ b/src/screenControl.py
@@ -513,15 +513,26 @@ class Controller(object):
             self.helperChrome.output(self.mode)
 
     def clearLine(self, y):
-        '''Clear a line of content, excluding the chrome'''
-        (minx, _, maxx, _) = self.getChromeBoundaries()
-        charsToDelete = range(minx, maxx)
-        # we go in the **reverse** order since the original documentation
-        # of delchar (http://dell9.ma.utexas.edu/cgi-bin/man-cgi?delch+3)
-        # mentions that delchar actually moves all the characters to the right
-        # of the cursor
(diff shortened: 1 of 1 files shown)


src/screenControl.py
```diff
@@ -513,15 +513,26 @@ def processDirty(self):
...
-        # of delchar (http://dell9.ma.utexas.edu/cgi-bin/man-cgi?delch+3)
-        # mentions that delchar actually moves all the characters to the right
-        # of the cursor
-        for x in reversed(charsToDelete):
-            self.stdscr.delch(y, x)
+        '''Clear a line of content, including chrome to the right, but not
+        to the left of the content window'''
+        # Q. Why do we have to delete the entire line?
+        # A. Because delchar deletes the character and moves the characters
+        # left on the line (to the right of the x spot) leftwards.
```
pcottle: `to the right of the cursor` might be better

pcottle:
Hrm, I'm going to hold off on merging this since the Travis CI results came back negative (seems like the tests actually failed).

Couldnt we just keep the right-to-left behavior though and just extend it to the edge of the chrome? that would consume the **whole** chrome content and then start chewing on the line, so the chrome doesnt bleed in.

lemme try that..

PR #128, opened 2015-05-22
clear also the chrome too, resolves #127

pcottle:
Yeah this seems a bit more minimal @lastquestion, gonna merge if thats ok with you :D


lastquestion:
Lol you don't like the 20 line comment I wrote? Haha. Yah both commits do the same thing, so I'm cool w either 👍

PR #131, opened 2015-05-22
Allow disabling file checks

pcottle:
Hopefully this is the last command line option we allow, but I found the filesystem checks a little constraining when working with deleted files that I want to restore (since they are obviously matches but we throw them away).

So here we add a bool param option and connect it up to the same bool param we use in tests.

screen shot 2015-05-22 at 10 13 31 am
screen shot 2015-05-22 at 10 13 25 am

### DIFF
--- a/src/processInput.py
+++ b/src/processInput.py
@@ -20,9 +20,11 @@ from usageStrings import USAGE_STR
 from screenFlags import ScreenFlags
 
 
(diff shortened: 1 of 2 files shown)


pcottle:
(The above is a before / after, where now we can highlight the deleted file)

PR #132, opened 2015-05-22
Resolves #130 Allow pluses in filenames

### LINKED ISSUE #130: Objective-C category conventions break path parsing
### Description
Objective-C categories are named like so: "NSArray+Utils.h". Path picker doesn't parse these properly and won't let you pick the entire path leading to the file. It only highlights the part after the "+". This creates significant usability issues while using fpp with Objective-C projects.

### Reproduce
In terminal:
```
$ touch NSArray+Utils.h
$ ls | fpp
```

pcottle:
We already allow `@` signs for retina images, and this surprisingly doesnt break any of our previous tests so lets go for it!

From some cursory investigation it also doesnt look like this introduces many false positives since we have filesystem validation now :clap: so much easier to expand our regex

### DIFF
--- a/src/__tests__/testParsing.py
+++ b/src/__tests__/testParsing.py
@@ -209,6 +209,23 @@ fileTestCases = [{
     'match': True,
     'num': 42,
     'file': './inputs/annoying Spaces Folder/evilFile With Space2.txt',
+}, {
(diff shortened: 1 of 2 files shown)


lastquestion:
:+1:

PR #134, opened 2015-05-22
Resolves #133 fix command mode

### LINKED ISSUE #133: Weird command mode highlighting bug
Once I select a file and go into command highlighting mode, it seems to highlight the whole area in the last color we painted with. @lastquestion did you notice this? Somewhat weird

screen shot 2015-05-22 at 10 34 03 am
screen shot 2015-05-22 at 10 34 01 am

pcottle:
One liner -- weird that this didnt get picked up in test coverage @lastquestion. Maybe when you dont specify an attribute to `self.stdscr.addstr` it doesnt actually reset the attribute back to 0? not sure. anyways, easy fix

### DIFF
--- a/src/screenControl.py
+++ b/src/screenControl.py
@@ -392,6 +392,7 @@ class Controller(object):
                 if index == self.hoverIndex]
 
     def showAndGetCommand(self):
+        self.stdscr.attrset(0)
         fileObjs = self.getFilesToUse()
         files = [fileObj.getFile() for fileObj in fileObjs]
         (maxy, maxx) = self.getScreenDimensions()


lastquestion:
Ooohh, we should fix this better by using the `printer` instead of `stdscr`: addstr in that implementation uses default color https://github.com/facebook/PathPicker/blob/master/src/colorPrinter.py#L53

actually using 0 is also not good, because it's white and i think on a bright console we'd get white on white? I didn't test yet cuz I haven't merged master today

pcottle:
Oh yeah good point, let me switch this

PR #135, opened 2015-05-22
Awesomeeee long file truncation Resolves #118

### LINKED ISSUE #118: Flag/configuration to shorten deep nested paths

pcottle:
Check the test coverage -- think this is pretty sweet! Also have gif demo:

http://g.recordit.co/Yk8xuJEXfv.gif


pcottle:
ugh apparently python3 is more restrictive about indices :*( thank you travis CI

pcottle:
@lastquestion ^^ might want to see the implementation of this. its pretty convenient!

PR #137, opened 2015-05-24
X Mode (rebased)

xavierbeynon:
Rebased previous x mode + nits + small test updates.


lastquestion:
I think there's a very small bug where if you're in X mode and you enter command mode, you'll be back in select mode when you are back on the main choose screen. Also I think the color of the text is white only because we are printing the scrollbar chrome right before, so the default color is set. It's not necessary but would be very good to use the color printer to affirmatively default to the default foreground color. (whew). I could do that in my cleanup PR #139 if @pcottle merges this first.

pcottle:
> I think there's a very small bug where if you're in X mode and you enter command mode, you'll be back in select mode when you are back on the main choose screen. 

yeah I could see that happening -- we probably need to track some kind of `lastMode` variable to keep track of where we came from

> Also I think the color of the text is white only because we are printing the scrollbar chrome right before, so the default color is set. It's not necessary but would be very good to use the color printer to affirmatively default to the default foreground color. (whew).

yeah do you mind doing this @xavierbeynon  and possibly adding a test for X-mode as well? It should be easy if you just provide `x` as one of the mocked inputs to the test case and check for the right screen. that way we can see what this looks like

src/screenControl.py
```diff
@@ -370,6 +381,8 @@ def processInput(self, key):
             # before exiting the program
             self.getFilesToUse()
             self.cursesAPI.exit()
+        elif self.mode == X_MODE and key in lbls:
```
pcottle: this might prone to breakage in the future since we might add keys above this statement that are in `lbls` and forget about `X_MODE`. i wonder if we could instead refactor some of this...

src/screenControl.py
```diff
@@ -561,3 +575,23 @@ def moveCursor(self):
...
+        self.mode = X_MODE if self.mode != X_MODE else SELECT_MODE
+        self.printAll()
+
+    def printXMode(self):
+        if self.mode == X_MODE:
+            (maxy, _) = self.scrollBar.screenControl.getScreenDimensions()
+            topY = maxy - 2
+            minY = self.scrollBar.getMinY() - 1
+            for i in range(minY, topY + 1):
+                self.stdscr.addstr(i, 1, lbls[i - minY])
```
pcottle: yeah using color printer here would be nice

xavierbeynon:
Sorry for being late guys - I'm between situations right now.  I could probably write a test this weekend and refactor the keys thing.  I don't really want to do anything with color printer since I don't know what that is.

pcottle:
> Sorry for being late guys - I'm between situations right now. I could probably write a test this weekend and refactor the keys thing. I don't really want to do anything with color printer since I don't know what that is.

its cool! let me see if i can merge this all in

pcottle:
Alright sweet! Thanks @xavierbeynon for the PR, rebasing was quite trivial (just a merge conflict in an expected test output file).

Now that I'm playing around with this locally it actually is quite fun / useful :)

we should try to figure out how to turn off the cursor in x mode -- one more thing to fix

pcottle:
Fixed! Alright I think this is all wrapped up now, thanks again :tada: :clap: this should make it into the 0.6.1 release

Issue #125, opened 2015-05-21
zsh issue with command mode


raine:
Happens when I write git add in the command mode:

```
$ git status | fpp
/Users/rainevi/.zprezto/init.zsh: line 14: autoload: command not found
/Users/rainevi/.zprezto/init.zsh: line 15: print: command not found
/Users/rainevi/commands.zsh: line 1: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
/Users/rainevi/commands.zsh: line 2: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
/Users/rainevi/commands.zsh: line 3: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
/Users/rainevi/commands.zsh: line 15: `mongo-cli': not a valid identifier
```

pcottle:
Hrm -- how do you source your zshrc? does sourcing these files work?

https://github.com/facebook/PathPicker/blob/master/src/output.py#L33

raine:
I'm not sure about the specific, but I use [prezto](https://github.com/sorin-ionescu/prezto)

I source `.zshsecrets` from `.zshrc` and `.zshsecret` sources `commands.zsh`

But the weird thing about that error output is that `commands.zsh` is not in the home directory, but rather in a directory under it.

pcottle:
Hrmmm. If you have a fresh bash script, can it successfully source your `.zshrc` file? something like

```
source ~/.zshrc
echo "success"
```

If you paste that into a `test.sh` and run it, what is the output? im wondering if its something weird with pretzo's configuration setup

raine:
I get the same errors when I do that.

pcottle:
Merged #138 so this is fixed! thanks again @hlian

pcottle:
whoops, this didnt post earlier:
=====

> I get the same errors when I do that.

yeah, so this is an issue with sourcing your zshrc to begin with -- one of the assumptions of `fpp`'s bash execution mode is that we can source all those files :P

let me check out #138 which is probably the better fix

raine:
Awesome.

PR #138, opened 2015-05-25
Use an interactive shell to run .fpp.sh

hlian:
```
    use interactive shell to execute .fpp.sh

    This conveniently executes all the appropriate rc files for both zsh and
    bash, thus expanding out aliases. There are cons to this approach
    though:

    * Though the Bash documentation strongly suggests that expand_aliases is
      turned on when interactive, I was not able to reproduce this behavior
      on Yosemite bash. (It's possible Yosemite just comes with a weird
      setup though.)

    * $SHELL only ever expands out to the user's default shell and not the
      current shell. (So in the rare case where you might open up zsh, then
      run bash, and then use fpp, fpp will choose zsh and not bash.)

    * Not all shells support the -i flag. However, the two shells currently
      supported by fpp (zsh and bash) do indeed have this handy dandy flag.

    A controversial change, perhaps! My feelings would not be hurt if you
    guys chose to not accept this PR, but my heart would be ecstatic if you
    did.

```

This fixes issue #125.

### DIFF
--- a/fpp
+++ b/fpp
@@ -30,7 +30,7 @@ function doProgram {
 
   $PYTHONCMD "$BASEDIR/src/choose.py" "$@" < /dev/tty
   # execute the output bash script
-  sh ~/.fpp/.fpp.sh < /dev/tty
+  $SHELL -i ~/.fpp/.fpp.sh < /dev/tty
 }
 
 # we need to handle the --help option outside the python
--- a/src/output.py
+++ b/src/output.py
@@ -29,12 +29,6 @@ versions which cannot be resolved.
 
 CONTINUE_WARNING = 'Are you sure you want to continue? Ctrl-C to quit'
 
-FILES_TO_SOURCE = [
-    '~/.zshrc',
-    '~/.bashrc',
-    '~/.bash_profile',
-    '~/.bash_aliases'
-]
 
 # The two main entry points into this module:
 #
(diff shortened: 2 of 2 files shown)


pcottle:
Hey @hlian! Thanks for the PR -- I remember interactive mode coming up when I was figuring out alias expansion but I think I got stuck on a similar issue you had -- yosemite needing `shopt` to turn on alias expansion yet that not being an option in zshrc. also stack overflow seemed to widely discourage use of interactive mode, but I think our use case is a valid exception.

I guess checking for the existence of shopt and then flipping the flag isn't the worst solution, but it does seem a *little* blunt (compared to sourcing).

Let me pull this down and test locally -- I might actually prefer this though since sometimes sourcing bashrc's can have pretty aggressive side effects (I personally have a script to login automatically to my facebook dev server if a given flag is present, so that always trips me up when using fpp with bash sourcing)

pcottle:
@frantic do you remember any reasons why we avoided interactive mode in the past? I think the `shopt` was the main reason but given how others might be running into #125 my preference is to merge this in.

Let's do it!

frantic:
Definitely cleaner than my attempt #59. Yeah, `shopt` I think, also bash on osx was weird.

(P.S. My hack to get `$SHELL` by invoking ps is so lame)

PR #139, opened 2015-05-25
Small cleanup

lastquestion:
* make all printing output use the `ColorPrinter` so we don't get any more bugs with wrong colors being used
* make fpp quit out immediately if `process` exits with error; this stops double printing of incorrect argument help 
* update command mode error screen test with attributes to catch regression fixed in #134 
* slight cleanup in `processInput`

### DIFF
--- a/fpp
+++ b/fpp
@@ -25,6 +25,9 @@ PYTHONCMD="python"
 function doProgram {
   # process input from pipe and store as pickled file
   $PYTHONCMD "$BASEDIR/src/processInput.py" "$@"
+  # if it failed, just fail now and exit the script
+  # this works for the looping -ko case as well
+  if [[ $? != 0 ]]; then exit $?; fi
   # now close stdin and choose input...
   exec 0<&-
 
--- a/src/formattedText.py
+++ b/src/formattedText.py
@@ -8,7 +8,7 @@
 import re
 import curses
 from collections import namedtuple
-
+from colorPrinter import ColorPrinter
 
 class FormattedText(object):
 
@@ -81,7 +81,8 @@ class FormattedText(object):
             if index % 2 == 1:
                 # text
                 toPrint = val[0:maxLen - printedSoFar]
-                printer.screen.addstr(y, x + printedSoFar, toPrint)
+                printer.addstr(y, x + printedSoFar, toPrint,
+                               ColorPrinter.CURRENT_COLORS)
                 printedSoFar += len(toPrint)
             else:
                 # formatting
--- a/src/processInput.py
+++ b/src/processInput.py
@@ -23,8 +23,7 @@ from screenFlags import ScreenFlags
 def getLineObjs(flags):
     inputLines = sys.stdin.readlines()
     return getLineObjsFromLines(inputLines,
-                                validateFileExists=False if
-                                flags.getDisableFileChecks() else True)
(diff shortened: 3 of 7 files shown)


fpp
```diff
@@ -25,6 +25,9 @@ PYTHONCMD="python"
 function doProgram {
   # process input from pipe and store as pickled file
   $PYTHONCMD "$BASEDIR/src/processInput.py" "$@"
+  # if it failed, just fail now and exit the script
+  # this works for the looping -ko case as well
+  if [[ $? != 0 ]]; then exit $?; fi
```
pcottle: ah nice -- i never figured out how to exit with the same status code as the previous command!

I bet we could probably remove the code that clears out the bash script as well after this change...

fpp
```diff
@@ -25,6 +25,9 @@ PYTHONCMD="python"
...
   # now close stdin and choose input...
   exec 0<&-
```
pcottle: github comments are lame and wont let me write below, but should we also have the same for `$BASEDIR/src/choose.py`? its less likely to fail but it might be handy for things like when the user quits and doesnt want to run the shell script. let me file a followup for that

src/formattedText.py
```diff
@@ -81,7 +81,8 @@ def printText(self, y, x, printer, maxLen):
             if index % 2 == 1:
                 # text
                 toPrint = val[0:maxLen - printedSoFar]
-                printer.screen.addstr(y, x + printedSoFar, toPrint)
+                printer.addstr(y, x + printedSoFar, toPrint,
+                               ColorPrinter.CURRENT_COLORS)
```
pcottle: ahh nice. what do you think about a sugar method like `printer.addCurrentColorStr`?

src/processInput.py
```diff
@@ -69,7 +68,6 @@ def usage():
             if os.path.isfile(filePath):
                 os.remove(filePath)
         print('Done! Removed %d files ' % len(stateFiles.getAllStateFiles()))
-        sys.exit(0)
```
pcottle: why is this gone? i think clean mode was specifically supposed to run separately and then exit early -- its not supposed to continue

src/formattedText.py
lastquestion: I like it, let's do it

pcottle:
:tada: :clap: thanks @lastquestion! let me check out the behavior of `--clean` real quick after this -- i might add the exit back

fpp
lastquestion: Yeah, that's a great idea. And haha I didn't know either, stackoverflow knows all.

lastquestion:
Woops, I didn't test `--clean`.yeah. thanks!

src/processInput.py
lastquestion: It's because I suck and I forgot to test it. My bad. thanks for catching it

fpp
pcottle: Followup from #131 -- @lastquestion do you know why this works for `processInput.py` but not `choose.py`? I'm assuming its because we connect to `dev/tty` or from closing stdin or something

lastquestion: Yeah i saw your comment on the other issue. `$?` _should_ work even tho we piping in input. I secretly suspect curses wrapper is maybe catching or rethrowing or something and the exit code is being lost..?

PR #140, opened 2015-05-25
truncate at the printing abstraction level

lastquestion:
This way we don't have to rebuild the `decoratedMatch` every time the screen size changes.

Additionally,
* fix a slight off by one rounding error in the original implementation that would leave one space empty on even sized input.
* add a test for when the screen is so small we can't even print truncation `|---|`


lastquestion:
#139 will cause this to merge conflict I think. I'll rebase after it gets merged.

src/formattedText.py
```diff
@@ -87,6 +87,72 @@ def printText(self, y, x, printer, maxLen):
...
+        else:
+            # no truncation
+            self.printText(y, x, printer, maxLen)
+            return
+
+        printedSoFar = 0
+        xIndex = x
+
+        for index, val in enumerate(self.segments):
+            if index % 2 == 1:
```
pcottle: whats this for? why not just do the specific segments we want?

lastquestion: the formatting and text are interleaved like this
`['formatting', 'text', 'formatting', 'text']`
so this selects either the formatting or text

pcottle: Ahh yeah forgot. Aight I'll take a closer look at this later -- its a lot of introduced code (that i dont fully grok yet) just to fix the off by one error :P

how does it handle the screen resizes better? Im actually wondering if its more or less performant, since the version earlier kind of cached whatever mode it was in

pcottle:
Yay! glad the off-by-one error is fixed and we handled the super narrow window case.

Any reason why you went with a larger change here @lastquestion? Im worried this might be a slight perf regression since now we are doing this calculation on every repaint rather than just when our status of needing truncation changes from false to true (or opposite).

Just worried about merging in a bunch of code that I know works but I dont quite grok yet :O

pcottle:
Closing this out since its stale, reopen if you want!

PR #143, opened 2015-05-29
Jump to line on Sublime Text

dufferzafar:
Add support for jumping to a specific line in sublime text.

The editor can be `subl` or `sublime` etc.


pcottle:
:tada: :clap: great PR! I missed this when switching from Facebook dev servers (which we all work on through ssh) to local-dev land.

sublime is super popular -- wonder if we should look up support for textmate or atom or anything else

PR #151, opened 2015-06-04
Update FPP github web styles

engz:
Already incorporated what we've discussed. One thing left is the copy to clipboard thingy. :dancer:

### DIFF
--- a/assets/script.js
+++ b/assets/script.js
@@ -1,0 +1,10 @@
+$(window).resize(function() {
+  console.log('resize called');
+  if ($(window).width() < 768) {
+    $("nav").addClass("navbar-fixed-top");
+  } else {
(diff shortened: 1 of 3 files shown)


assets/script.js
```diff
@@ -0,0 +1,10 @@
+$(window).resize(function() {
+  console.log('resize called');
+  if ($(window).width() < 768) {
+    $("nav").addClass("navbar-fixed-top");
+  } else {
+    $("nav").removeClass("navbar-fixed-top");
+  }
+});
+
+resize();
```
pcottle: ill probably just inline this JS rather than have a whole separate file (which is another web request)

pcottle:
:tada: :heart_eyes_cat: thank you so much @engz ! Let me merge this in

engz:
:tada:

PR #156, opened 2015-06-05
after selecting a file go to the next one

prubel:
I never find myself wanting to stay on the same file,
and ofter want to select a group of files. Simplify that
job.

### DIFF
--- a/src/screenControl.py
+++ b/src/screenControl.py
@@ -372,6 +372,7 @@ class Controller(object):
             self.jumpToIndex(self.numMatches - 1)
         elif key == 'f':
             self.toggleSelect()
+            self.moveIndex(1)
         elif key == 'A' and not self.mode == X_MODE:
             self.toggleSelectAll()
         elif key == 'ENTER':


pcottle:
Hrmmm... if we had //launched// with this feature I think it could have gone in, but now that we've been released for quite some time I don't want to break everyones muscle memory by adding this :crying_cat_face: itd be pretty thrashy to update and then have all your normal keystrokes lead to unexpected output.

I also think each keystroke should be fairly "KISS" or keep-it-simple-stupid. Although usually the intention is to go downwards after selecting a file, the cases where that isnt true ends up being quite thrashy. Imagine you have a really long list and you actually only want to select the last entry in the file -- by moving the index to +1, you'll scroll all the way to the beginning of the list -- not even giving you an option to see if the selection worked. that would feel like a bug.

additionally say you are selecting files on the way up (rather than downwards) -- then you have to fight against each selection pushing you down.

This also breaks almost every test we have (since the expected output is different from keystrokes), hence why the build failed.

Anyways, I'm going to close this out since it has some pros but also some cons. Thanks of the interest in the project though @prubel !

prubel:
Good points. I should have run the tests. What about something like 'F'? or maybe 'd' for down, and it's also right next to f?

pcottle:
Hrm, I could be convinced for F! that way its a separate keystroke but similar (select-and-down)

yeah if you run the tests and add another for the F functionality we could merge this in

PR #157, opened 2015-06-05
update docs to show how to run tests.

prubel:
The previous file no longer exists, or at least I couldn't find it


pcottle:
Ah of course! my bad @prubel -- we updated the travis formula and everything but I forgot the docs. :tada: :clap: thanks

PR #159, opened 2015-06-10
add new 'F' command to select and go down.

prubel:
I haven't changed the usage yet in narrow mode as that will touch
many of the tests and that bottom is getting kind of wide as it is. 

Do we need other tests for this?

I was thinking that some inline help in narrow mode via h might help
alleviate the wide bottom issue. I've got some code, but it's not all the way polished.

### DIFF
--- a/src/__tests__/expected/dontWipeChrome.txt
+++ b/src/__tests__/expected/dontWipeChrome.txt
@@ -2,39 +2,39 @@
  ______________________                                  G                                                                                                                                               
  fpp                                                |  4 +|    * [f] toggle the selection of a file
                                                          G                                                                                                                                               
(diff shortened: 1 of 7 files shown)


prubel:
Gack, I'll get the test fixed up and resubmit. CI sure is nice.

pcottle:
This looks great! yeah I bet all the tests pass without the usage string change, but a lot of them have the usage in it so youll have to delete and re-run the test. the git diff ends up showing the real changes at the end of the day

this looks great though! :clap:

pcottle:
Thanks for the update @prubel !! :clap: :tada: :tada: this looks great!

PR #164, opened 2015-06-19
Resolves #163 resolve just file with spaces

### LINKED ISSUE #163: Can not handle files with spaces in their filenames. 0.6.1
Use "ls -la | fpp" in a directory and if the directory has files which names have spaces, fpp can not get the right file names. So these files are unselectable(as shown in the attached image
0335e4d2-a790-497b-a919-285aa17f5f98
)

I checked the source code and found that in parse.py, only MASTER_REGEX_WITH_SPACES has the ability to parse filename like "a b c.txt". But this regex can only parse string with back slash like "/hello/world/a b c.txt". When I execute "ls", the only thing fpp can have is filename without any "/".

I think this can be solved in two ways:
1. change MASTER_REGEX_WITH_SPACES from "(([a-z.A-Z0-9\-_]|\s[a-zA-Z0-9\-_])+\/)+" to "(([a-z.A-Z0-9\-_]|\s[a-zA-Z0-9\-_])+\/)*"
2. call prependDir() before call matchLine()

pcottle:


pcottle:
#yolo gonna merge this in since it passes locally

PR #170, opened 2015-06-24
Fix .deb Lintian Error (ManPage Compression)

pallavagarwal07:
Note that the file needs to be compressed, with the original deleted, and not as a copy.


pcottle:
I trust you! :clap:

sethwoodworth:
Thank you! Although I run Debian, I'd never added a manpage to a Debian package before.

PR #172, opened 2015-07-07
Point to the latest release

evverx:


pcottle:
`/me facepalm`

Thanks @evverx! I was meaning to make the README updating part of the script that runs when we build a new release, but I had no idea that GitHub releases has a `/Latest/` auto redirect. how convenient! this is much better

Issue #55, opened 2015-05-08
dotfiles are ignored


mellowcandle:
for instance if I have 
.gitignore
in my git status output. it can't be selected using PathPicker.

metavida:
I've got a WIP fix for this: https://github.com/metavida/PathPicker/tree/include_git_status_dotfiles

pedrox:
I believe dot files inside directory are not a problem.
So the issue is actualy about dot files and with not period in the current directory. Do you confirm? It seems that those are not detected by design since it's more likely that they are not paths in the input.

pedrox:
This is a duplicate of #7.

metavida:
Thanks for checking out this ticket @pedrox. First, I'd like to make a clarification. When I say "dotfile" I mean "any file that begins with a period"[[1]].

> I believe dot files inside directory are not a problem.

I confirmed that PathPicker does **not** recognize dotfiles within a directory by adding the following case to `test.py` on the latest master branch commit.

```py
{
    'input': 'tmp/.gitignore',
    'match': True,
    'file': 'tmp/.gitignore',
    'num': 0
}
```
```bash
$ git log -n1 | head -n1
commit 7b0bef202b9f4783197459e26fe658f404b99188
$ python src/test.py
...
======================================================================
FAIL: testFileMatch (__main__.TestParseFunction)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "src/test.py", line 244, in testFileMatch
    self.checkFileResult(testCase)
  File "src/test.py", line 252, in checkFileResult
    testCase['input'])
AssertionError: Line "tmp/.gitignore" did not match any regex

----------------------------------------------------------------------
Ran 5 tests in 0.007s

FAILED (failures=2, errors=1)
```

> This is a duplicate of #7.

To my reading, issue #7 is about detecting files with no extensions (e.g. `Makefile`). While this issue is similar, I would suggest that this ticket is *not* a duplicate of #7 because it's quite possible to have a dotfile with an extension (e.g. `.env.production` or `.hidden.txt`).

[1]:http://en.wikipedia.org/wiki/Dot-file

PR #173, opened 2015-07-22
Detect current directory [A-Z]*file, eg. Makefile

Shenil:
I see that matching files without extensions in the current directory has already been discussed at length before, such as in https://github.com/facebook/PathPicker/issues/83 and https://github.com/facebook/PathPicker/issues/55. 

Feel free to close if this is too narrow a use-case, but I find matching for current directory [A-Z]*files, a la Podfile, Gemfile, Makefile, etc. to be a useful addition because these files often appear in the root directories for projects, and would not create too many false positives.

### DIFF
--- a/src/__tests__/testParsing.py
+++ b/src/__tests__/testParsing.py
@@ -273,6 +273,17 @@ fileTestCases = [{
     'match': True,
     'disableFuzzTest': True,
     'file': 'evilFile No Prepend.txt',
+}, {
+    'input': 'Gemfile',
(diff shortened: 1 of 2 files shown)


pcottle:
Whoops sorry I missed this @Shenil ! I agree its already been discussed a bunch but since this PR has test coverage and covers a pretty valid usecase (files often in the root directories), I'm more than happy to merge it.

Thanks for contributing!

Shenil:
Anytime. Glad to see it get merged!

PR #178, opened 2015-08-15
Line number support for vi, neovim, nano and atom.

pallavagarwal07:
Added rest of the common editors for the sake of completion.


pcottle:
:tada: :heart_eyes: thanks @pallavagarwal07 !!

Issue #185, opened 2015-08-26
Issuing command within vim triggers suspend of vim


alecjacobson:
I'm trying to use `fpp` from within `vim`. For example, in vim I pipe the output of `ls` into `fpp`

```bash
:!ls | fpp
```

This works as usual bringing up the familiar `fpp` interface. Suppose I choose a file `f` and then enter command mode with `c` and enter `echo `. This successfully echos the selected filename but has an untended side effect of suspending my vim session. I'm back to my bash terminal and see

```bash
[1]+  Stopped      vim
```

pcottle:
Just a random question first -- what functionality of `fpp` are you looking for over something like `NERDTree`? the multiple selection and command mode perhaps?

Anyways, I'm actually quite surprised this works as much as you described :P So the command gets correctly echo-ed out, but the vim session never resumes?

I'm worried it's this line:
`$SHELL -i ~/.fpp/.fpp.sh < /dev/tty`

can you possibly try without the `< /dev/tty` hook? should be able to edit the `fpp` bash script easily:
```
[pcottle:~:]$ type -a fpp
fpp is /usr/local/bin/fpp
```

alecjacobson:
Exactly, I like the multiple selection + command mode. (I hadn't heard of NERDTree before).

Removing `< /dev/tty` from that line did not fix the problem, but removing the `-i` did fix it (though I'm not sure what that could now have broken).

pcottle:
> but removing the -i did fix it (though I'm not sure what that could now have broken).

gah, yeah then we might have to close this out unless we can detect if we are inside the vim shell or not.

basically the `-i` option forces it to be interactive, so things like opening up vim / emacs / `git add -p` / whatever will actually work. Seems like vim's execution mode gets tripped up by running an interactive script.

Seems like theres a similar issue over here too:
https://github.com/zaiste/vimified/issues/89

pcottle:
(for context, we need the `-i` to both expand bashrc aliases iirc and make the tty stuff work. so reverting it would be a fairly big regression.

im open to ideas on running conditionally with the `-i` option if we can somehow tell we are inside a vim subshell)

PR #187, opened 2015-08-31
-i flag only if not in vim shell

alecjacobson:
Hopefully an answer to the problem in https://github.com/facebook/PathPicker/issues/185 (-i flag in shell command causing vim to be suspended).

### DIFF
--- a/fpp
+++ b/fpp
@@ -32,8 +32,13 @@ function doProgram {
   exec 0<&-
 
   $PYTHONCMD "$BASEDIR/src/choose.py" "$@" < /dev/tty
+  # Determine if running from within vim shell
+  IFLAG=""
+  if [ -z "$VIMRUNTIME" ]; then
+    IFLAG="-i"
+  fi
   # execute the output bash script
-  $SHELL -i ~/.fpp/.fpp.sh < /dev/tty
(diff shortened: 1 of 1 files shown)


pcottle:
Ah sweet! Does this fix #185 for you? Didn't realize we had something like `$VIMRUNTIME`

keep in mind that I think bashrc aliases wont expand in this mode (without the `-i` option) but I would need to check. thats better than hanging the shell though.

pcottle:
Another thought -- should we prevent going into editor mode if we are already inside a vim runtime when we launch `fpp`? I dont think theres a valid case for running vim as launched from a vim bash command

alecjacobson:
I agree it's better than hanging the shell. It's also possible to load
.bashrc etc. explicitly using the source command.

Personally, I actually _do_ like the ability to launch the editor from vim
mode. This matches the behavior of issue `:vim filename` within vim: a new
vim process is run, quitting that one will pop up to the original one.

pcottle:
Lets do it! Thanks for putting up the fix -- this will make the next release, but in the meantime you can tell your friends to install from HEAD on master (since you already have the fix locally)

PR #192, opened 2015-09-26
Fix issue with long screens

slackorama:
My screen had more rows than the size of `lbls` so when I chose quick
select on a full list of candidates, I got a traceback "IndexError:
string index out of range."

This only displays labels up to the number of labels and no more.


pcottle:
Thanks @slackorama so much for putting up the PR for this! logic seems good -- its too bad we don't have test coverage for this though. I'll try to add something after this lands.

Also I should make a new release since this and a few other fixes have gone in recently... lemme file a task on myself to do that

Issue #193, opened 2015-09-28
Doesn't jump to line number when using neovim


jeffkile:
Since I upgrade vim to neovim fpp stopped jumping to the line number of the selected item, instead it just opens the file and stops at the first line.

pcottle:
Hey @jeffkile do you know the command line formatting for `neovim`'s line number args? We have a manual mapping here:
https://github.com/facebook/PathPicker/blob/371ffc2d8df002e16e7e063ce218023280fab50d/src/output.py#L95

and neovim has an entry. does that not work? can you output ` ~/.fpp/.fpp.sh`?

jeffkile:
Yeah that manual mapping is wrong. It should be this
```
    if editor in ['vim', 'nvim'] and lineNum != 0:
        return '\'%s\' +%d' % (filePath, lineNum)
    elif editor in ['vi', 'nano', 'joe', 'emacs'] and lineNum != 0:
        return '+%d \'%s\'' % (lineNum, filePath)
```

Can I submit a pull request?

pcottle:
certainly! Would be much appreciated :) cc @pallavagarwal07 just as an fyi. i shoulda double checked in diff review though

pallavagarwal07:
That's weird. I've been using neovim for over 6 months and it supports both
`nvim +17 'file.c'`
and
`nvim 'file.c' +17` formats.

pcottle:
hrm, might be version dependent then? @jeffkile what version are you running?

jeffkile:
`NVIM 0.0.0-alpha+201508100001 (compiled Aug 11 2015 10:53:00)` 
I'm using using Fish, tmux and OSX as well.

I just tried it from the command line and oddly enough both ways @pallavagarwal07 suggested work fine for me ... but only from the command line. When using FPP only the line number after the file name works for me.

Should I still push up the PR or should I look in to this further?

pcottle:
Have you outputted your `$FPP_EDITOR` or `$VISUAL`? we use those environment variables before falling back to `$EDITOR`. that might be the difference in behavior

or output the `~/.fpp/.fpp.sh` and paste it here so we can compare exactly

pallavagarwal07:
@jeffkile Could you ensure that the file is being opened in nvim (`:version` after opening file via fpp)? It might be not obvious if both vim and nvim use the same vimrc and maybe it's reading nvim environment variable and thus nvim format but attempting to open the file finally with vim?

pcottle:
[...]

PR #195, opened 2015-10-02
Fix issue where emacsclient didn't go to line

slackorama:
Add emacsclient to the list of editors for the linenumber match.


pcottle:
sometimes i wish python had switch statements... thanks for the awesome PR :+1: . @jeffkile is working on changing the neovim mapping over in #193

PR #196, opened 2015-10-03
Remove unneeded conditional in vim handling

ismith:
I believe the handling of 'vim foo' and 'vim -O foo' (the single-file case) are identical, and so this conditional is unnecessary.  (Not sure, though.  I was in here because I found the use of -O a bit surprising.  Worked around by setting `FPP_EDITOR='vim '`; it might be good to call out in the docs that the default behavior has that flag and can be worked around, but I'm not sure where to put it.)

### DIFF
--- a/src/output.py
+++ b/src/output.py
@@ -116,10 +116,7 @@ def expandPath(filePath):
 def joinEditCommands(partialCommands):
     editor, editor_path = getEditorAndPath()
     if editor in ['vim', 'mvim']:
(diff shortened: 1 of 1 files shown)


pcottle:
Yeah I think I was just being overly cautious here, seems fine to me. thanks for the negative LOC PR :+1: 

i think it'd be worth checking in neovim, but it does seem odd to throw for that combination

PR #197, opened 2015-10-05
[SHELL] Delegate to bash for all other shells Resolves #186

### LINKED ISSUE #186: Doesn't work when login shell is csh
Darwin wkoszek-macbook.local 14.5.0 Darwin Kernel Version 14.5.0: Wed Jul 29 02:26:53 PDT 2015; root:xnu-2782.40.9~1/RELEASE_X86_64 x86_64

My login shell is CSH. To rule out CSH, I've run Bash in iTerm2 window and tested too. Same problem. I tested in Vagrant with ubuntu/trusty64. Stuff works OK there, so what I describe below seems to be MacOSX specific.

I did:

brew update
brew install fpp.

cd /tmp
mkdir sample
cd sample
cal 2010 > 2010
cal 2011 > 2011
cal 2012 > 2012

/bin/ls -1 ./* | fpp

I select 2010, press f, select 2011, press f, press c, type 'cat', <ENTER>. Nothing happens. I'm dropped to fpp subshell. I'd expect to see output from cat'ed files.

---------------- cat ~/.fpp/.fpp.sh ----------------------
if type shopt > /dev/null; then
  shopt -s expand_aliases
fi

echo "executing command:"
echo "cat './2010' './2011'"
cat './2010' './2011'
-----------------------------------------------------------------

So selection part works OK. Execution is doing something wrong.

I started to add some debugging code: https://github.com/wkoszek/PathPicker/

pcottle:


pcottle:
@frantic do you mind taking a look at this? #186 has the context (just look at the tail end)

fpp
```diff
@@ -37,8 +37,18 @@ function doProgram {
...
+  # shells, we delegate to $SHELL, but for all others
+  # (fish, csh, etc) we delegate to bash.
+  #
+  # We use the following heuristics from
+  # http://stackoverflow.com/questions/3327013/
+  # in order to determine which shell we are on
+  if [ -n "$BASH" -o -n "$ZSH_NAME" ]; then
+    $SHELL $IFLAG ~/.fpp/.fpp.sh < /dev/tty
+  else
+    /bin/sh $IFLAG ~/.fpp/.fpp.sh < /dev/tty
```
frantic: this might not work on Linux. IIRC on Mac `/bin/sh` is bash.

pcottle: i looked at http://askubuntu.com/questions/141928/what-is-difference-between-bin-sh-and-bin-bash briefly, but yeah now that i look closer apparently `/bin/sh` on some distros is "dash" not bash. ill switch :P

PR #198, opened 2015-10-07
use /bin/bash instead of /sh/

pcottle:
#197 for context, and  http://askubuntu.com/questions/141928/what-is-difference-between-bin-sh-and-bin-bash


pcottle:
@frantic :P

frantic:
:fireworks:

PR #205, opened 2015-10-25
Remove $? from exit status for Fish Shell compatibility

benmccormick:
This solves the problems mentioned in #186 and doesn't change behavior in bash/zsh since `exit;` is equivalent to `exit $?;`.

### DIFF
--- a/src/output.py
+++ b/src/output.py
@@ -195,7 +195,7 @@ def appendToFile(command):
 
 
 def appendExit():
-    appendToFile('exit $?;')
+    appendToFile('exit;')
 
 
 def writeToFile(command):


pcottle:
Lets do it! We should probably also remove the logic that defers to either `zsh` or `bash` over in `fpp`

pcottle:
https://github.com/facebook/PathPicker/blob/master/fpp#L47

Want to add a case for fish there and check if it works? thanks again @benmccormick for your contribution! After we get this settled I'll probably add a new release

benmccormick:
@pcottle so the problem is that fish users already fall into that if statement (which is why this was a problem to begin with)

Since fpp is a bash script (`#!/bin/bash` at the top), it gets run in bash even when called from fish.  That causes $BASH to be defined, but $SHELL to still be fish.  If $FISH had been hitting the second if statement and we were running the script with bash, everything would have been fine before as long as I didn't try to use any fishy commands.  

As is, with this fix, everything should work fine and I can use fish scripting commands if I want.

I don't believe the second half of that if statement is ever going to be hit though, unless csh/ksh/whatever act differently when running a bash script.  If you want what I think was the intended behavior, you'll have to check whether $SHELL ends with bash, fish, or zsh.  Personally I think you're probably just better off axing the if statement though.

benmccormick:
1__uiworkspace___users_ben_code_windsor_wcui_dev_src_wcui__tmux_

Quick example of the behavior described in my last comment.

pcottle:
> As is, with this fix, everything should work fine and I can use fish scripting commands if I want.

Ah great, thanks for the explanation then. Also sorry for the delay, was some craziness prepping for black friday over at FB.

#209 mentions formally supporting `fish` as a shell. do you think adding an if case condition for fish would be worthwhile? the you could run fishy commands right?

benmccormick:
@pcottle I personally don't do very much with Fish as a scripting language, so it doesn't matter to me.  I use fish for the autocomplete/history and write my scripts in python for the most part, with an occasional fish or bash script that I use shebangs to run with the correct shell.  I don't care about doing fancy fishy things within PathPicker, 90% of the time I'm just using it to open files in an editor.  What I cared about with this was it not crashing if I did `ack -l foo | fpp`

Other people might care more, but its not a big deal to me

PR #210, opened 2015-11-23
Exposing "--all-input" option.

gsheld:
A new feature for `fpp` that would allow the user to take full advantage of the input selection GUI without being constrained to input with file paths.

For one small example, one could apply `fpp` to git branches:
```
# checkout selected branch
git branch | fpp -ai -c "git checkout --"

# delete selected branch
git branch | fpp -ai -c "git branch -D"
```

In "all input" mode, all lines are taken as valid input, assuming they contain more than just whitespace; each line is also trimmed of whitespace at its beginning and end.

### DIFF
--- a/src/__tests__/testParsing.py
+++ b/src/__tests__/testParsing.py
@@ -339,6 +339,32 @@ prependDirTestCases = [
         'out': ''
     }]
 
+allInputTestCases = [
+    {
+        'input': '    ',
+        'match': None
+    }, {
+        'input': ' ',
+        'match': None
+    }, {
+        'input': 'a',
+        'match': 'a'
+    }, {
+        'input': '   a',
+        'match': 'a'
+    }, {
+        'input': 'a    ',
+        'match': 'a'
+    }, {
+        'input': '    foo bar',
+        'match': 'foo bar'
+    }, {
+        'input': 'foo bar    ',
+        'match': 'foo bar'
+    }, {
+        'input': 'foo bar baz',
+        'match': 'foo bar baz'
+    }]
 
 class TestParseFunction(unittest.TestCase):
 
@@ -397,6 +423,23 @@ class TestParseFunction(unittest.TestCase):
         for testCase in fileTestCases:
             self.checkFileResult(testCase)
         print('Tested %d cases.' % len(fileTestCases))
+
+    def testAllInputMatches(self):
+        for testCase in allInputTestCases:
+            result = parse.matchLine(testCase['input'], False, True)
+
+            if not result:
+                self.assertTrue(testCase['match'] is None,
+                                'Expected a match "%s" where one did not occur.' %
+                                testCase['match'])
+
+                continue
+
+            (match, _, _) = result
+            self.assertEqual(match, testCase['match'], 'Line "%s" did not match.' %
+                            testCase['input'])
+
+        print('Tested %d cases for all-input matching.' % len(allInputTestCases))
 
     def checkFileResult(self, testCase):
         result = parse.matchLine(testCase['input'],
--- a/src/screenControl.py
+++ b/src/screenControl.py
@@ -35,11 +35,17 @@ X_MODE = 'X_MODE'
 
 lbls = "ABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890~!@#$%^&*()_+<>?{}|;'"
 
-SHORT_NAV_USAGE = '[f|A] selection, [down|j|up|k|space|b] navigation, [enter] open, [x] quick select mode, [c] command mode'
+# options for displayed to the user at the bottom of the screen
+SHORT_NAV_OPTION_SELECTION_STR = '[f|A] selection'
+SHORT_NAV_OPTION_NAVIGATION_STR = '[down|j|up|k|space|b] navigation'
+SHORT_NAV_OPTION_OPEN_STR = '[enter] open'
+SHORT_NAV_OPTION_QUICK_SELECT_STR = '[x] quick select mode'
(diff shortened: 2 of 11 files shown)


src/screenFlags.py
```diff
@@ -92,6 +95,13 @@ def getArgParser():
 of, say, deleted files in git status that you would like to restore to a given
 revision. It enables you to select the deleted files even though they
 do not exist on the system anymore.''')
+        parser.add_argument('-ai',
+                            '--all-input',
+                            default=False,
+                            action="store_true",
+                            help='''You may force PathPicker to recognize all
+lines as acceptible input. Typically, PathPicker will scan the input for references
+to file paths. Passing this option will disable those scans and the program will assume that every input line is a match.''')
```
pcottle: could add an explanation that its helpful for things like git branches, mercurial bookmarks, maybe even directories -- who knows!

pcottle:
This is absolutely awesome! Really appreciate the thorough test coverage and cleanup along the way @gsheld. I'll take a deeper look tomorrow with some nits, but this can likely go in with minimal change.

I agree that combining with a preconfigured command mode is nice. maybe provide a few example commands for git to isolate the bookmarks? for example, mercurial outputs something like:

```
[pcottle:~/www:(explainNoMethods)]$ hg book
   betterStats.preamend      1095557:c0e02bfc109b
   debugSF                   1098767:f13654075612
   emailRoundups             1095558:6cd95011bcf3
 * explainNoMethods          1102007:d620f805c859
```

so if we matched on all text, you couldn't pipe it straight into `hg update`. `git branch` is a bit better but technically the currently checked out branch will have the `*` :P

also might have been nice to splice the `file` => `path` codemod into another PR, but nbd. im pretty lazy when interchanging the two, even though they arent technically equivalent

src/__tests__/testParsing.py
```diff
@@ -339,6 +339,32 @@
...
+        'match': 'a'
+    }, {
+        'input': '    foo bar',
+        'match': 'foo bar'
+    }, {
+        'input': 'foo bar    ',
+        'match': 'foo bar'
+    }, {
+        'input': 'foo bar baz',
+        'match': 'foo bar baz'
```
pcottle: i know this would pass, but maybe add a test case with whitespace on both sides, just for completeness

```
input: '   foo bar  ',
match: 'foo bar',
```

src/__tests__/testParsing.py
```diff
@@ -398,6 +424,23 @@ def testFileMatch(self):
...
 
+    def testAllInputMatches(self):
+        for testCase in allInputTestCases:
+            result = parse.matchLine(testCase['input'], False, True)
+
+            if not result:
+                self.assertTrue(testCase['match'] is None,
+                                'Expected a match "%s" where one did not occur.' %
+                                testCase['match'])
+
```
pcottle: nit: rm this line for readability, since its hard to tell what block continue is for

src/screenControl.py
```diff
@@ -115,14 +122,27 @@ def outputBottom(self):
...
         self.printer.addstr(borderY + 1, self.getMinX(), usageStr)
 
+    def getShortNavUsageString(self):
+        navOptions = [SHORT_NAV_OPTION_SELECTION_STR,
+                      SHORT_NAV_OPTION_NAVIGATION_STR,
+                      SHORT_NAV_OPTION_QUICK_SELECT_STR,
+                      SHORT_NAV_OPTION_COMMAND_STR]
+
+        # it does not make sense to give the user the option to "open" the selection
+        # in all-input mode
```
pcottle: ahhh, i see why you factored this out into separate strings now. cool!

src/screenControl.py
```diff
@@ -115,14 +122,27 @@ def outputBottom(self):
...
+        if not self.flags.getAllInput():
+            navOptions.insert(2, SHORT_NAV_OPTION_OPEN_STR)
```
pcottle: the insert 2 is weird here, i think could do something like

```
return ', '.join([
  SELECT_STR,
  NAV_STR,
  # comment
  None if self.flags.getAllInput() else NAV_OPEN_STR,
].filter(something()?));
```

but thats just me trying to be more functional. dont know how hard that is in practice with the filter, havent written python in a while

pcottle:
Alright more or less looked over everything -- some inline suggestions and nits that you can do in a followup (ideally), but this can go in! thanks again

gsheld:
@pcottle Glad you liked the diff. I'll create another PR for cleanup when I have a chance. I actually found a small bug with the patch you just merged as well, so that will also be addressed in the next PR.

Thanks for reviewing!

src/screenControl.py
gsheld: Yes, its not very intuitive. I think this approach works best; its more or less what you're suggesting.

```
    def getShortNavUsageString(self):
        navOptions = [SHORT_NAV_OPTION_SELECTION_STR,
                      SHORT_NAV_OPTION_NAVIGATION_STR,
                      SHORT_NAV_OPTION_OPEN_STR,
                      SHORT_NAV_OPTION_QUICK_SELECT_STR,
                      SHORT_NAV_OPTION_COMMAND_STR]

        # it does not make sense to give the user the option to "open" the selection
        # in all-input mode
        if self.flags.getAllInput():
            navOptions.remove(SHORT_NAV_OPTION_OPEN_STR)

        return ', '.join(navOptions)
```

pcottle: ah yeah, removing is a bit cleaner imo. it just felt like happing to map the index with 0-offset back to the original array to figure out where it would be inserted was a bit tough to read. but a super small minor nit of course

PR #214, opened 2015-12-14
Fix fatal

pcottle:


pcottle:
passed locally so I'm assuming this is good

PR #216, opened 2015-12-15
change git clone

joanna-chen:
the cloning command from before brought up a permission error for me
this correction works


pcottle:
Ah yes, because you might not have permissions :) great point! merging now, thanks a bunch

PR #218, opened 2016-01-11
Clean up Debian package

ssgelm:
This fixes two issues I had with the debian package:

 * **Remove package.sh when running dpkg so that it doesn't get included in the resulting package**
    
    If not removed the `dpkg --build` command will add package.sh to the system root which is not desired.
 * **Change the output name of the package to follow the debian standard**

    As described at https://www.debian.org/doc/manuals/debian-faq/ch-pkg_basics.en.html the package should be of the format of `<foo>_<VersionNumber>-<DebianRevisionNumber>_<DebianArchitecture>.deb`

I would have done two PRs but these changes seemed pretty simple so it seemed like overkill.  Let me know if you'd prefer that.  CLA has been signed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,5 @@
 *.orig
 dist/*
 fpp.deb
+fpp*.deb
 debian/usr/
--- a/debian/package.sh
+++ b/debian/package.sh
@@ -41,11 +41,12 @@ find . -type d -exec chmod 755 {} \;
 find . -type f -exec chmod 644 {} \;
 
 echo "Building package..."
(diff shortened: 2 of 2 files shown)


pcottle:
Thanks a bunch! @pallavagarwal07 do you mind just taking a quick sanity check look? We have pallavagarwal07 to thank for all the debian support, even if it wasn't 100% perfect :)

The explanation and code seems straightforward here though, so think this should be safe to merge.

pallavagarwal07:
I never noticed the first mistake. Thanks a bunch for fixing that. Seems great !

PR #221, opened 2016-01-15
Recognise 'vim -p' as vim

niteria:
My $EDITOR is vim -p and it'd be awesome if it jumped to line numbers just like regular vim.


pcottle:
thanks. double the commit points for a single change :)

Issue #206, opened 2015-11-02
Sending editor to background gets stuck


Firehed:
If you send your editor to the background (via `^z`) when it was opened by FPP, the whole thing gets stuck apparently waiting for a signal and won't end until you kill FPP via `kill` (`^c` is unresponsive), rather than returning to the shell as usual. So you get something like this:

    10:56:32 firehed@Erics-MacBook-Pro ~/some/path [0]$ grep -r www *|fpp
    (do things in vim, background with ctrl-z)
    3 files to edit
    ^C^C^C
    ^Z
    fg
    (`kill $FPP_PID` from another window)
    Vim: Caught deadly signal HUP
    Vim: Finished.

This is when using Vim on OS X in Bash, FPP v0.6.2 (installed via Homebrew). It didn't happen in the previous version I had installed; unfortunately, I have no idea what version that was (but it was something near the first public release).

pcottle:
Yeah, this is unfortunately due to the fact that we launch the editor inside a subshell and attach it to tty directly. hence you cant simply suspend the editor process, since it belongs in that sub-shell process :(

I think there was a similar issue to this earlier -- theres no real way to solve this i dont think, especially with supporting both zsh and bash.

thanks for using FPP though! Lemme know if you have ideas (im not an expert on launching processes and separating them from the script being run, but also still blocking on them before exiting)

pcottle:
Good news @Firehed  we are adding a flag for this in #222

PR #222, opened 2016-01-19
add a command-line flag to force non-interactive mode

ngavalas:
The change (https://github.com/facebook/PathPicker/pull/138) to add the -i flag to the subshell that we use to execute .fpp.sh causes some breakages when it comes to using I/O and job control.  If you launch vim, for example, from fpp, you can't Ctrl+Z to suspend vim because of interactive mode; it completely hangs the terminal.  

Interactive mode is not really meant to execute shell scripts the way we're using it in fpp, but since it's already in and I don't want to really break some users' workflows, I'd like to add a flag to spawn the subshell with a normal bash invocation.  I don't need aliases, but I'd really like job control and basic ideas around output redirection to work properly.

### DIFF
--- a/fpp
+++ b/fpp
@@ -21,6 +21,7 @@ done
 BASEDIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 
 PYTHONCMD="python"
+NONINTERACTIVE=false
 
 function doProgram {
   # process input from pipe and store as pickled file
@@ -34,7 +35,7 @@ function doProgram {
   $PYTHONCMD "$BASEDIR/src/choose.py" "$@" < /dev/tty
   # Determine if running from within vim shell
   IFLAG=""
-  if [ -z "$VIMRUNTIME" ]; then
+  if [ -z "$VIMRUNTIME" -a "$NONINTERACTIVE" = false ]; then
     IFLAG="-i"
   fi
   # execute the output bash script. For zsh or bash
@@ -67,6 +68,8 @@ for opt in "$@"; do
     exit 0
   elif [ "$opt" == "--record" -o "$opt" == "-r" ]; then
     echo "Recording input and output..."
+  elif [ "$opt" == "--non-interactive" -o "$opt" == "-n" ]; then
+    NONINTERACTIVE=true
   elif [ "$opt" == "--keep-open" -o "$opt" == "-ko" ]; then
     # allow control-c to exit the loop
     # http://unix.stackexchange.com/a/48432
--- a/fpp
+++ b/fpp
@@ -68,7 +68,7 @@ for opt in "$@"; do
     exit 0
   elif [ "$opt" == "--record" -o "$opt" == "-r" ]; then
     echo "Recording input and output..."
-  elif [ "$opt" == "--non-interactive" -o "$opt" == "-n" ]; then
+  elif [ "$opt" == "--non-interactive" -o "$opt" == "-ni" ]; then
     NONINTERACTIVE=true
   elif [ "$opt" == "--keep-open" -o "$opt" == "-ko" ]; then
     # allow control-c to exit the loop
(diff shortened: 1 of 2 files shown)


fpp
```diff
@@ -67,6 +68,8 @@ for opt in "$@"; do
     exit 0
   elif [ "$opt" == "--record" -o "$opt" == "-r" ]; then
     echo "Recording input and output..."
+  elif [ "$opt" == "--non-interactive" -o "$opt" == "-n" ]; then
```
pcottle: i feel like `-n` is a bit short here. `-ni`? just in case we need n for something later

pcottle:
Thanks for putting this up @ngavalas ! I'll go find the other issue we had with interactive vim mode and let them know we are adding a flag.

Travis CI seems to have caught something:

```
  File "testScreen.py", line 15, in <module>
    import screenTestRunner
  File "/home/travis/build/facebook/PathPicker/src/__tests__/screenTestRunner.py", line 16, in <module>
    import choose
  File "../choose.py", line 17, in <module>
    import screenControl
  File "../screenControl.py", line 13, in <module>
    import processInput
  File "../processInput.py", line 19, in <module>
    from usageStrings import USAGE_STR
  File "../usageStrings.py", line 9, in <module>
    from screenFlags import ScreenFlags
  File "../screenFlags.py", line 49
    return self.args.non_interactive
                                   ^
TabError: inconsistent use of tabs and spaces in indentation
Tests failed :*(
The command "./scripts/buildAndTest.sh" exited with 1.
```

can you try fixing it up and running the script again locally?

src/screenFlags.py
```diff
@@ -45,6 +45,9 @@ def getDisableFileChecks(self):
     def getAllInput(self):
         return self.args.all_input
 
+    def getIsNonInteractive(self):
```
pcottle: this is not used but i guess we can keep it here for consistency

pcottle:
:tada: !! And our contributor count goes up by one haha. feel free to send a diff for the internal version so you dont have to wait for me to sync it over

ches:
Hi, sorry for commenting/asking a question on a closed issue, but this seems the most relevant place. I can open a new one if you like.

I came here from #206 because I just hit that issue, and it caused me to lose work in my editor. That's pretty nasty default behavior, having not noticed that this `--non-interactive` flag exists until now (it's only helpful after you know you need it 😢 ).

> Interactive mode is not really meant to execute shell scripts the way we're using it in fpp, but since it's already in and I don't want to really break some users' workflows

I agree, and I'd argue that this bad experience and possibility of losing work warrants considering changing the default, maybe when a major version bump comes along if that's the project's policy? It's still `0.x` so if you're following Semantic Versioning then all bets are still off… I'd be curious to hear use cases where people actually depend on the *interactive* shell behavior, beyond aliases, etc. which I think can be solved.

This is from the bash man page (emphasis mine):

> When  bash is invoked as an interactive login shell, **or as a non-inter-
> active shell with the --login option**, it first reads and executes  com-
> mands  from  the file /etc/profile, if that file exists.  After reading
> that file, it looks for ~/.bash_profile, ~/.bash_login, and ~/.profile,
> in  that order, and reads and executes commands from the first one that
> exists and is readable.

If loading aliases and such is desired, I believe that's the "right" way to do it: a login shell, but not interactive.

I can try to work up a change to propose and see how it works, but before digging in I wanted to ask, @pcottle said this on #206:

> this is unfortunately due to the fact that we launch the editor inside a subshell and attach it to tty directly

Is there a reason or some background you can point me to on why fpp doesn't just `exec` the process?

Thanks for the tool, it's useful!

pcottle:
> Is there a reason or some background you can point me to on why fpp doesn't just exec the process?

I'm not a bashrc expert at all, but if you just blindly "exec" the process from either python or the wrapper bash script, it doesnt attach to TTY so your editor runs in the background with no one interacting with it.

the reason why we have all these hacks is because we want to go from:

  - piped input
  - interactive screen (TTY)
  - then either a straight command to execute OR another interactive screen (vim emacs or whatever)

Issue #226, opened 2016-04-04
Command mode not working with fish


uri:
To reproduce

```
git status | fpp
F
c
ls
<return>
```

This produces the following error

```
Missing end to balance this if statement
~/.fpp/.fpp.sh (line 3): if type shopt > /dev/null; then
                         ^
fish: Error while reading file /Users/uri/.fpp/.fpp.sh
```

My `$SHELL` env var is set to `/usr/local/bin/fish`

#### versions

- fish: `fish, version 2.2.0`
- fpp: `fpp version 0.7.0`

pcottle:
Can you try adding some fish detection to this line:
https://github.com/facebook/PathPicker/blob/master/src/output.py#L173

? Seems like zsh doesnt mind that block, but fish does

uri:
Yeah removing the call to `appendToFile` fixes the problem.

petr-tik:
is this closed then?

pcottle:
I think so! Sorry for not updating sooner

PR #227, opened 2016-04-04
Fixes command mode for fish

uri:
Not running appendAliasExpand when fish is set as shell.

See #226 for more details.

### DIFF
--- a/src/output.py
+++ b/src/output.py
@@ -170,7 +170,9 @@ def appendAliasExpansion():
     # despite documentation hinting otherwise.
     #
     # so here we must ask bash to turn on alias expansion.
-    appendToFile("""
+    shell = os.environ.get('SHELL')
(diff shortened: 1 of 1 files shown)


src/output.py
```diff
@@ -170,7 +170,9 @@ def appendAliasExpansion():
     # despite documentation hinting otherwise.
     #
     # so here we must ask bash to turn on alias expansion.
-    appendToFile("""
+    shell = os.environ.get('SHELL')
+    if 'fish' not in shell:
```
pcottle: couldhave just done `if fish in shell: return` but nbd

pcottle:
Thanks for the fix!!

src/output.py
uri: Good to know for next time!

PR #228, opened 2016-04-05
Add support for 'Page Up' and 'Page Down'

robertbachmann:
Use curses keys 'PPAGE' and 'NPAGE' instead of
'PAGE_UP' and 'PAGE_DOWN'


src/charCodeMapping.py
```diff
@@ -13,8 +13,6 @@
 CODE_TO_CHAR.update((value, name[4:]) for name, value in vars(curses).items()
                     if name.startswith('KEY_'))
 # special exceptions
-CODE_TO_CHAR[4] = 'PAGE_DOWN'
```
pcottle: do you know what keys these are? been a while since i wrote this

src/screenControl.py
```diff
@@ -383,9 +383,9 @@ def processInput(self, key):
             self.toggleXMode()
         elif key == 'c':
             self.beginEnterCommand()
-        elif key == ' ' or key == 'PAGE_DOWN':
```
pcottle: great, i wanted to make sure we kept the space and b keys

pcottle:
AHhhhh I swear I googled this and never got it to work. thanks so much, especially for adding the test! :tada:

pcottle:
i might whip up a release in a few days, since this is pretty nice to have

src/charCodeMapping.py
robertbachmann: On Linux 4 seems to be CTRL+D
Don't know about 20.

pcottle:
New release is live!
https://github.com/Homebrew/homebrew-core/pull/370

PR #229, opened 2016-04-05
Add suport for 'Home' and 'End' keys

robertbachmann:
- 'Home' selects the first file
- 'End' selects the last file


pcottle:
🎉 👍  sounds great to me

PR #231, opened 2016-06-02
Allow use of FPP_DIR environment variable

cryptostrophe:
Use the value of the environment variable `FPP_DIR` (if present) instead of `~/.fpp`.

(This is my first pull request. I have completed the CLA under this GitHub username.)

### DIFF
--- a/fpp
+++ b/fpp
@@ -23,10 +23,6 @@ BASEDIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 PYTHONCMD="python"
 NONINTERACTIVE=false
 
-if [ -z "$FPP_DIR" ]; then
-  FPP_DIR=~/.fpp
-fi
-
 function doProgram {
   # process input from pipe and store as pickled file
   $PYTHONCMD "$BASEDIR/src/processInput.py" "$@"
--- a/fpp
+++ b/fpp
@@ -23,6 +23,10 @@ BASEDIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 PYTHONCMD="python"
 NONINTERACTIVE=false
 
+if [ -z "$FPP_DIR" ]; then
+  FPP_DIR=~/.fpp
+fi
+
 function doProgram {
   # process input from pipe and store as pickled file
   $PYTHONCMD "$BASEDIR/src/processInput.py" "$@"
--- a/src/stateFiles.py
+++ b/src/stateFiles.py
@@ -9,7 +9,7 @@ from __future__ import print_function
 
 import os
 
-FPP_DIR = '~/.fpp'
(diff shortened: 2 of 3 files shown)


fpp
```diff
@@ -23,6 +23,10 @@ BASEDIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 PYTHONCMD="python"
 NONINTERACTIVE=false
 
+if [ -z "$FPP_DIR" ]; then
+  FPP_DIR=~/.fpp
```
pcottle: why add this at all? we already have the default down on read (which is the better one imo). then we dont need to have two different sources of truth or places to update it

src/stateFiles.py
```diff
@@ -9,7 +9,7 @@
 
 import os
 
-FPP_DIR = '~/.fpp'
+FPP_DIR = os.environ.get('FPP_DIR') or '~/.fpp'
```
pcottle: 👍

pcottle:
Can we add the fix I outlined above? I think just modifying the python is sufficien!

fpp
cryptostrophe: Probably an overabundance of caution (this is actually my first PR ever). I will add another commit without the shell-script change!

pcottle:
> Probably an overabundance of caution (this is actually my first PR ever).

No worries at all, welcome to the club! Glad you chose PathPicker with the honor 👍 

This looks great now, i'll go ahead and merge. Not sure when our next release is, we kinda just had one recently, but it will go out with the next release!

edentsai:
When I use `$FPP_DIR` environment variable in bash, i got a problem below:

``` bash
[~/demo-fpp] $ mkdir my-fpp
[~/demo-fpp] $ ls -a my-fpp/
./  ../
[~/demo-fpp] $ export FPP_DIR="$HOME/demo-fpp/my-fpp"

[~/demo-fpp] $ echo 'hello' > foo
[~/demo-fpp] $ echo 'hello' > foo2
[~/demo-fpp] $ find foo* -type f -d 0 | fpp
bash: ~/.fpp/.fpp.sh: No such file or directory

[~/demo-fpp] $ ls -a my-fpp/
./        ../       .fpp.log  .fpp.sh   .pickle
```

Because [fpp script](https://github.com/facebook/PathPicker/blob/master/fpp#L48) still try to read file from `~/.fpp/.fpp.sh`
```
  if [ -n "$BASH" -o -n "$ZSH_NAME" ]; then
    $SHELL $IFLAG ~/.fpp/.fpp.sh < /dev/tty
  else
    /bin/bash $IFLAG ~/.fpp/.fpp.sh < /dev/tty
  fi
```

edentsai:
I send a PR to fixes above:
https://github.com/facebook/PathPicker/pull/232

PR #234, opened 2016-07-23
added test case for describe file feature

rjdean123:
is there a better way to add a change to a PR that was already merged and closed?

anyways, here's a passing test case, let me know if that's not what you were looking for 😄

### DIFF
--- a/src/__tests__/expected/testDescribeFile.txt
+++ b/src/__tests__/expected/testDescribeFile.txt
@@ -1,0 +1,60 @@
+src/charCodeMapping.py         | 20 +++++++++++++
+BBBBBBBBBBBBBBBBBBBBBB                                                          
+src/choose.py                  | 15 ++++++++--
+_____________                                                                   
+src/colorPrinter.py            | 21 ++++++++-----
+___________________                                                             
+src/cursesAPI.py               | 40 +++++++++++++++++++++++++
+________________                                                                
+src/format.py                  |  4 +--
+_____________                                                                   
+src/processInput.py            |  7 +++++
+___________________                                                             
+/foo/bar/src/charCodeMapping.py         | 20 +++++++++++++
+                                                                                
+/foo/bar/src/choose.py                  | 15 ++++++++--
+                                                                                
+/foo/bar/src/colorPrinter.py            | 21 ++++++++-----
+                                                                                
+/foo/bar/src/cursesAPI.py               | 40 +++++++++++++++++++++++++
+                                                                                
+/foo/bar/src/format.py                  |  4 +--
+                                                                                
+/foo/bar/src/processInput.py            |  7 +++++
+                                                                                
+/foo/bar/src/screenControl.py           | 28 +++++++-----------
+                                                                                
+/foo/bar/src/screenFlags.py             | 34 +++++++++++++++++++++
+                                                                                
+14 files changed, 290 insertions(+), 33 deletions(-)
+                                                                                
+
+                                                                                
+
+                                                                                
+
+                                                                                
+
+                                                                                
+
+                                                                                
+
+                                                                                
+
+                                                                                
+
+                                                                                
+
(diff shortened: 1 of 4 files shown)


pcottle:
its almostttt what we want -- the issue is that since the screen is by default narrow, the sidebar doesnt show, so your command output doesnt show. thats what im trying to get test coverage over

i think you want to add
```
 'screenConfig': {
         'maxX': 201,
         'maxY': 40
     },
```
to your test case. then delete the file and re-run the test and make sure the output file as your command description area

rjdean123:
gah, failing because the 'last accessed' line is different each run, will look more into this when I get some time tomorrow

pcottle:
Oh dangit yeah, that's tricky. That must mean that doing the ls and cat commands on the file ends up recording it as accessed.

If that's the case, is it worth having that line in the output? Its basically always going to be the current day / second if describing the file accesses it, right?

rjdean123:
yeah, that's what i figured--removed the time last accessed feature. luckily i dont think that was one of the more useful features, I think time last modified is more likely to be helpful anyways

pcottle:
```
  File "testScreen.py", line 209, in testScreenInputs
    allInput=('-ai' in args or '--all-input' in args),
  File "/home/travis/build/facebook/PathPicker/src/__tests__/screenTestRunner.py", line 62, in getRowsFromScreenRun
    choose.doProgram(screen, flags, CursesForTest(), lineObjs)
  File "../choose.py", line 42, in doProgram
    screen.control()
  File "../screenControl.py", line 359, in control
    self.processInput(inKey)
  File "../screenControl.py", line 443, in processInput
    self.describeFile()
  File "../screenControl.py", line 349, in describeFile
    self.helperChrome.outputDescription(self.lineMatches[self.hoverIndex])
  File "../screenControl.py", line 77, in outputDescription
    self.outputDescriptionPane(lineObj)
  File "../screenControl.py", line 122, in outputDescriptionPane
    lineObj.getOwnerUser(),
  File "../format.py", line 130, in getOwnerUser
    output = subprocess.check_output(bashCommand.split())
AttributeError: 'module' object has no attribute 'check_output'
----------------------------------------------------------------------
Ran 1 test in 0.912s
FAILED (errors=1)
Tests failed :*(
```

hrm, seems like the tests fail now. mind checking? so close!

rjdean123:
ah, of course haha. weird, they passed locally before I pushed. 

will take a look later today.

rjdean123:
so the expected output file is being generated for a file on my machine, and when it's tested on Travis, the information doesn't line up, which is actually correct behavior. is there a way to generate the expected output on the test server?

But, luckily digging through this issue, I did also realize that a python function i was using was new to python 2.7, and so I was able to rewrite some things to make sure fpp plays nicely with 2.6+/3.0+.

pcottle:
ooooh ugh i know why. its because the files are "new" when they get slurped over to travis and probably have different ownerships or whatever, so the expected values are different

there are two options:

- use dependency injection / mocking to stub out the data returned. thats harder probably
- add a simple `ignoreLines:` parameter to the test config that just ignores comparing the passed in lines:

```
ignoreLines: [45, 49, 51, 53]
```

that way we still execute the code, we just dont care what those specific lines are -- since they will change all the time

sorry about the trouble here! i just want to double check your code is executing and tested -- even if we dont compare the output, getting it executing means if it throws a fatal we will catch that.

and it sounds like it caught the 2.7/3.0 issue already :)

rjdean123:
Got it, yeah, the tests have already proven to be helpful!

taking a look at this now.

rjdean123:
apologies this is taking so long, got quite busy at the end of the summer/start of the new year. should be good to work on this this weekend.

pcottle:
No worries!

pcottle:
closing out since its stale

PR #240, opened 2016-09-01
Fix unit tests

shish:
Looks like somebody committed some expected output without committing the input?


pcottle:
Thank you!! Was wondering why it passed locally but the build was failing

🎉 🎉

PR #252, opened 2016-10-23
Add autopep8 command detection to build script

Conky5:
Add detection for `autopep8` and fail the build if it is not available.
Add a blurb to `CONTRIBUTING.md` that it is a dependency as well.


Conky5:
Updated to allow travis to do the `autopep8` check and make formatting corrections so that the build passes.

scripts/buildAndTest.sh
```diff
@@ -1,4 +1,10 @@
 #!/bin/bash
+
+if ! type autopep8 > /dev/null; then
+  echo "'autopep8' required for build."
+  exit 1;
+fi
+
 PEPLINES=$(autopep8 --recursive ./src/ --diff | wc -l)
```
pcottle: so with my clowny code beforehand, what did this line do? since autopep8 was missing, it probably just wrote some output to stderr and then the counted lines of stdout was 0, so it skipped?

wow thats lucky :O

pcottle:
🎉 🎉 👍 

thanks a ton! this is great, commented inline on my guess at how this even worked beforehand but curious if you have any theories.

Conky5:
@pcottle you're welcome! You're correct, it wrote to stderr and "Pep8 Compliant!" would be printed, and the script would return success.

PR #253, opened 2016-12-01
refactored output.py to correct vim behavior

brwong:
- changed vim output from columns to tabs
- changed vim command so that jump-to-line commands work

(Feel free to modify as needed.)

### DIFF
--- a/scripts/makeDist.sh
+++ b/scripts/makeDist.sh
@@ -2,7 +2,7 @@
 VERSION="$(python ./src/version.py)"
 DEST="./dist/fpp.$VERSION.tar.gz"
 mkdir -p ./dist/
-tar -cf $DEST src/*.py fpp
+tar -czf $DEST src/*.py fpp
 
 sed -i '' -e "s#url .*#url \"https://github.com/facebook/PathPicker/releases/download/$VERSION/fpp.$VERSION.tar.gz\"#g" ./fpp.rb
 HASH=$(cat $DEST | shasum -a 256 | cut -d " " -f 1)
--- a/src/output.py
+++ b/src/output.py
@@ -48,7 +48,8 @@ def execComposedCommand(command, lineObjs):
 
 def editFiles(lineObjs):
     logger.addEvent('editing_num_files', len(lineObjs))
-    filesAndLineNumbers = [(lineObj.getPath(), lineObj.getLineNum()) for lineObj in lineObjs]
+    filesAndLineNumbers = [(lineObj.getPath(), lineObj.getLineNum())
+                           for lineObj in lineObjs]
     command = joinFilesIntoCommand(filesAndLineNumbers)
     appendIfInvalid(lineObjs)
     appendToFile(command)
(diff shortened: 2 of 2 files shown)


brwong:
My vim command reference, FYI

[Stack Overflow - Vim : Open multiples files on different lines](http://stackoverflow.com/questions/2148397/vim-open-multiples-files-on-different-lines/2148692#2148692)

pcottle:
Hrm, most people at work I talk to prefer columns rather than tabs (just because its immediately clear that `$x` files are open). And my other worry is that this has been the default behavior for over a year, and changing in this release would be a bit thrashy

would you be open to putting this behind some kind of system flag?

brwong:
I changed it so that "vim" acts as before (but with the correct line number jumps), and setting FPP_EDITOR to "vim -p" opens new tabs.

As usual, I'm open to suggestions.

scripts/makeDist.sh
```diff
@@ -2,7 +2,7 @@
 VERSION="$(python ./src/version.py)"
 DEST="./dist/fpp.$VERSION.tar.gz"
 mkdir -p ./dist/
-tar -cf $DEST src/*.py fpp
+tar -czf $DEST src/*.py fpp
```
pcottle: this is `czf` in trunk and was changed recently. you probably reverted this in some kind of rebase or something -- mind dropping the change to this file?

src/output.py
```diff
@@ -94,33 +92,36 @@ def getEditorAndPath():
...
+    if editor == 'vim -p':
+        firstFilePath, firstLineNum = filesAndLineNumbers[0]
+        cmd += ' +%d %s' % (firstLineNum, firstFilePath)
+        for (filePath, lineNum) in filesAndLineNumbers[1:]:
+            cmd += ' +"tabnew +%d %s"' % (lineNum, filePath)
+    elif editor == 'vim':
+        firstFilePath, firstLineNum = filesAndLineNumbers[0]
+        cmd += ' +%d %s' % (firstLineNum, firstFilePath)
+        for (filePath, lineNum) in filesAndLineNumbers[1:]:
+            cmd += ' +"vsp +%d %s"' % (lineNum, filePath)
```
pcottle: woahhh so `vsp` does vertical split like `-O` and allows you to specify the line numbers? very cool!

src/output.py
```diff
@@ -94,33 +92,36 @@ def getEditorAndPath():
     return 'vim', 'vim'
 
 
-def getEditFileCommand(filePath, lineNum):
```
pcottle: ah i see why this had to go -- yeah you didnt have global context when composing these mini-parts.

src/output.py
```diff
@@ -94,33 +92,36 @@ def getEditorAndPath():
...
-        return editor_path + ' -O ' + ' '.join(partialCommands)
-    # Assume that all other editors behave like emacs
-    return editor_path + ' ' + ' '.join(partialCommands)
+    cmd = editor_path + ' '
+    if editor == 'vim -p':
+        firstFilePath, firstLineNum = filesAndLineNumbers[0]
+        cmd += ' +%d %s' % (firstLineNum, firstFilePath)
+        for (filePath, lineNum) in filesAndLineNumbers[1:]:
+            cmd += ' +"tabnew +%d %s"' % (lineNum, filePath)
+    elif editor == 'vim':
```
pcottle: note to self -- we lost the `FPP_DISABLE_SPLIT` behavior here so lets add that back

PR #262, opened 2017-05-19
Allow preconfigured commands through custom key bindings

saveriomiroddi:
For use cases where PP is used frequently with the same command, it's quite annoying to have to always type the same command (eg. `rspec` for Ruby testing).

This PR allows the user to associate custom commands that can be executed through specified keys (eg. `r` for `rspec`), speeding up the PP workflow for repetitive command executions.

Custom bindings/commands are stored in the `<FPP_DIR>/.fpp.keys` (in the `[bindings]` group) as standard text configuration file. The existing clean FPP internal interface allows such configuration functionality to be trivially extended, for example, in case FPP will implement a static configuration.

This implementation is the simplest possible; notably, it doesn't support keys already bound (which would significantly complicate the feature).

Due to the test framework, it's not easy to write an end-to-end test - in fact, the existing functionality for executing command hasn't this type of tests; therefore, the test have been updated to inspect the visualization of the custom keys/commands.

### DIFF
--- a/scripts/buildAndTest.sh
+++ b/scripts/buildAndTest.sh
@@ -15,7 +15,7 @@ else
 fi
 
 cd ./src/__tests__/
-python testParsing.py && python testScreen.py > /dev/null
+python testParsing.py && python testScreen.py && python testKeyBindingsParsing.py > /dev/null
 if [ $? -eq 0 ]
 then
   echo "Tests passed!"
--- a/src/__tests__/keyBindingsForTest.py
+++ b/src/__tests__/keyBindingsForTest.py
@@ -1,0 +1,14 @@
+# Copyright (c) 2015-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+#
+
+KEY_BINDINGS_FOR_TEST_CONFIG_CONTENT = "[bindings]\nr = rspec\ns = subl\n"
+
+
+class KeyBindingsForTest(object):
+
+    bindings = [('r', 'rspec'.encode('utf-8')), ('s', 'subl'.encode('utf-8'))]
--- a/src/__tests__/screenTestRunner.py
+++ b/src/__tests__/screenTestRunner.py
@@ -20,6 +20,7 @@ from screenFlags import ScreenFlags
 
 from screenForTest import ScreenForTest
 from cursesForTest import CursesForTest
+from keyBindingsForTest import KeyBindingsForTest
 
 INPUT_DIR = './inputs/'
 
@@ -59,7 +60,8 @@ def getRowsFromScreenRun(
     # we run our program and throw a StopIteration exception
     # instead of sys.exit-ing
     try:
-        choose.doProgram(screen, flags, CursesForTest(), lineObjs)
+        choose.doProgram(screen, flags, KeyBindingsForTest(),
+                         CursesForTest(), lineObjs)
     except StopIteration:
         pass
 
--- a/src/choose.py
+++ b/src/choose.py
@@ -17,6 +17,7 @@ import screenControl
 import logger
 import format
 import stateFiles
+from keyBindings import KeyBindings
 from cursesAPI import CursesAPI
 from screenFlags import ScreenFlags
 
@@ -29,16 +30,19 @@ this error will go away)
 '''
 
 
-def doProgram(stdscr, flags, cursesAPI=None, lineObjs=None):
+def doProgram(stdscr, flags, keyBindings=None, cursesAPI=None, lineObjs=None):
     # curses and lineObjs get dependency injected for
     # our tests, so init these if they are not provided
+    if not keyBindings:
+        keyBindings = KeyBindings()
     if not cursesAPI:
         cursesAPI = CursesAPI()
     if not lineObjs:
         lineObjs = getLineObjs()
     output.clearFile()
     logger.clearFile()
-    screen = screenControl.Controller(flags, stdscr, lineObjs, cursesAPI)
+    screen = screenControl.Controller(
+        flags, keyBindings, stdscr, lineObjs, cursesAPI)
     screen.control()
 
 
--- a/src/keyBindings.py
+++ b/src/keyBindings.py
@@ -1,0 +1,42 @@
+# Copyright (c) 2015-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+#
+import os
+import sys
+
+if sys.version_info[0] < 3:
+    import ConfigParser
+    parserModule = ConfigParser
+else:
+    import configparser
+    parserModule = configparser
+
+from stateFiles import FPP_DIR
+
+KEY_BINDINGS_FILE = os.path.join(FPP_DIR, '.fpp.keys')
+
+
+class KeyBindings(object):
+
+    bindings = []
+
+    def __init__(self, keyBindingsFile=KEY_BINDINGS_FILE):
+        """Returns configured key bindings, in the format [(key, command), ...].
+        The ordering of the entries is not guaranteed, although it's irrelevant to the purpose.
+        """
+        configFilePath = os.path.expanduser(keyBindingsFile)
+        parser = parserModule.ConfigParser()
+        parser.read(configFilePath)
+
+        # The `executePreconfiguredCommand` underlying APIs use `curses.getstr()`, which returns an
+        # encoded string, and invoke `decode()` on it.
+        # In Python 3/configparser, the parsed strings are already decoded, therefore don't support
+        # this method anymore, so we convert them to encoded ones first.
+        #
+        if parser.has_section("bindings"):
+            self.bindings = [(key, command.encode('utf-8'))
+                             for key, command in parser.items("bindings")]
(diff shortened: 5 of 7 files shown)


saveriomiroddi:
Hello!

Is there any guideline for Python 3.x (in particular, related to the configuration parser import)?

I've noticed that just running `scripts/buildAndTest.sh` on my system (as instructed in `CONTRIBUTING.md`) is not enough - it needs to be run on all the Python supported versions.

pcottle:
thanks so much for taking a stab at this! cool to see people put up PR's for ideas :)

> I've noticed that just running scripts/buildAndTest.sh on my system (as instructed in CONTRIBUTING.md) is not enough - it needs to be run on all the Python supported versions.

yeah so its a bit annoying, since it just uses the python in the path variable. i have python 3.something installed as `python3` and just hack the script to use that instead (or set a higher priority path variable that uses that)

i see that you extend the command display interface to show each command that you have binded. a few ideas:
- this might not scale if people add dozens of commands, and since this is a power user feature anyways, why not perhaps just dont change the display at all?
- related to that, an alternative would be to have a single command that shows the list of binded commands (full screen) that you can then exit out of with a single keypress.

but yeah seems like we need to change to `configparser` in python 3+:
```
OK
  File "testScreen.py", line 15, in <module>
    import screenTestRunner
  File "/home/travis/build/facebook/PathPicker/src/__tests__/screenTestRunner.py", line 16, in <module>
    import choose
  File "../choose.py", line 20, in <module>
    from keyBindings import KeyBindings
  File "../keyBindings.py", line 2, in <module>
    import ConfigParser
ImportError: No module named 'ConfigParser'
Tests failed :*(
```

https://docs.python.org/3/library/configparser.html

src/__tests__/screenTestRunner.py
```diff
@@ -20,6 +20,7 @@
 
 from screenForTest import ScreenForTest
 from cursesForTest import CursesForTest
+from keyBindingsForTest import KeyBindingsForTest
```
pcottle: glad you found this hacky dependency injection jazz haha

src/choose.py
```diff
@@ -29,16 +30,19 @@
 '''
 
 
-def doProgram(stdscr, flags, cursesAPI=None, lineObjs=None):
+def doProgram(stdscr, flags, keyBindings=None, cursesAPI=None, lineObjs=None):
```
pcottle: should we make this required? we pass it in everywhere right?

src/keyBindings.py
```diff
@@ -0,0 +1,22 @@
...
+
+    bindings = []
+
+    def __init__(self):
+        """Returns configured key bindings, in the format [[key, command], ...]
+        """
+        configFilePath = os.path.expanduser(
+            os.path.join(FPP_DIR, KEY_BINDINGS_FILE))
+        parser = ConfigParser.ConfigParser()
+        parser.read(configFilePath)
```
pcottle: if the file doesnt exist what happens?

saveriomiroddi:
Related to the bindings display (can't reply to your reply):

> this might not scale if people add dozens of commands, and since this is a power user feature anyways, why not perhaps just dont change the display at all?
> related to that, an alternative would be to have a single command that shows the list of binded commands (full screen) that you can then exit out of with a single keypress.

OK; I think it's a good idea to go for the idea #1, then there's always time to implement #2 as extra feature - agile style :-)

saveriomiroddi:
Related to the configuration parser:

> but yeah seems like we need to change to configparser in python 3+:

I will propose some code for handling both versions; I don't think there is a pretty way of doing so, but let's see.

src/choose.py
saveriomiroddi: If I'm understanding well, this translates to `def doProgram(stdscr, flags, keyBindings, ...)`; I think it's a good idea.

Yes, the Controller (which is the "center" class) definitely needs it.

There is also the interesting, open, idea that the KeyBindings class (and its backing file) is in reality nothing else than a generic configuration class.
In the future, it will be very easy to convert it to a more general `Configuration` (or any other name).

Now, I'm saying this to strenghten the point that it's worth making required, being a configuration.

src/keyBindings.py
saveriomiroddi: Nothing, the code is tolerant :-) But it's good practice to add some tests; I will add a test suite.

saveriomiroddi:
OOOK, I've commented everything. I will update the PR with the proposed changes. It will take some time because I've never programmed in Python, and I will need to maintain a good quality :-) So please bear with me :-)

pcottle:
> OOOK, I've commented everything. I will update the PR with the proposed changes. It will take some time because I've never programmed in Python, and I will need to maintain a good quality :-) So please bear with me :-)

no problem at all!! this is a great first stab if you've never programmed in python; my apologies that you have to handle both python2 and python3, and the APIs are slightly different. maybe now you know why the transition to 3 has been stalled for years :P

good luck with the update, and yeah adding some tests would be nice but if its overly difficult I wouldnt stress about it. I can do some manual testing locally, and it seems like the codepath doesnt really change if you dont have this config file.

related to that -- can we only change the display of the hotkey bindings if you have a config file in place? that will prevent you from having to modify a bunch of test files

smiroddi:
Hey Peter!

The PR is completed and ready for review. Compared to the last commit, the (relevant) changes are:

- custom bindings are not shown anymore;
- added python 3 support;
- added test suite for the bindings configuration file decode.

The last doubt I have is whether the binding detection fragment (`for boundKey, command in self.keyBindings.bindings:`) should stay at the top or the bottom of `processInput()`.
If we assume this usage is advanced, I think it should stay at the top, since it gives the option to override standard fpp keys.

smiroddi:
Note that I've followed your indication:

> this might not scale if people add dozens of commands, and since this is a power user feature anyways, why not perhaps just dont change the display at all?

I'm not sure now about this:

> related to that -- can we only change the display of the hotkey bindings if you have a config file in place? that will prevent you from having to modify a bunch of test files

Before removing the display code with the current commit, nothing additional (compared to a configuration without bindings) was displayed when there were no bindings defined - am I misunderstanding?

scripts/buildAndTest.sh
```diff
@@ -15,7 +15,7 @@ else
 fi
 
 cd ./src/__tests__/
-python testParsing.py && python testScreen.py > /dev/null
+python testParsing.py && python testScreen.py && python testKeyBindingsParsing.py > /dev/null
```
pcottle: we should probably change this to `python test*` but this is ok for now

src/__tests__/keyBindingsForTest.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) 2015-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+#
+
+KEY_BINDINGS_FOR_TEST_CONFIG_CONTENT = "[bindings]\nr = rspec\ns = subl\n"
```
pcottle: ah interesting so you use this in the testing below but you define it here? guess thats nice to keep it in sync with the mock

src/__tests__/keyBindingsForTest.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class KeyBindingsForTest(object):
+
+    bindings = [('r', 'rspec'.encode('utf-8')), ('s', 'subl'.encode('utf-8'))]
```
pcottle: why do you have to encode to utf8? probably just for equality comparison right?

src/keyBindings.py
```diff
@@ -0,0 +1,42 @@
...
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+#
+import os
+import sys
+
+if sys.version_info[0] < 3:
```
pcottle: ah great!

src/keyBindings.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def __init__(self, keyBindingsFile=KEY_BINDINGS_FILE):
+        """Returns configured key bindings, in the format [(key, command), ...].
+        The ordering of the entries is not guaranteed, although it's irrelevant to the purpose.
+        """
+        configFilePath = os.path.expanduser(keyBindingsFile)
+        parser = parserModule.ConfigParser()
+        parser.read(configFilePath)
+
+        # The `executePreconfiguredCommand` underlying APIs use `curses.getstr()`, which returns an
```
pcottle: ahh i see, interesting

src/screenControl.py
```diff
@@ -461,7 +462,10 @@ def processInput(self, key):
             self.cursesAPI.exit()
         elif self.mode == X_MODE and key in lbls:
```
pcottle: cool, so X mode is still respected first with priority

pcottle:
> Before removing the display code with the current commit, nothing additional (compared to a configuration without bindings) was displayed when there were no bindings defined - am I misunderstanding?

nope, that comment isnt relevant anymore if you dont display the bindings at all :) I was just suggesting two options -- dont show the bindings at all, or only show them if you have a config file (and dont include a config file in the test cases).

it seemed like your original version of this PR added a bindings file for all the test cases, so all the test cases would have to be updated. but thats no longer relevant if we hide them :)

pcottle:
hrm `python test*.py` fails in the unittest console:
```
AttributeError: 'module' object has no attribute 'testParsing'
```

guess that was the only polish I was going to add to this, looks great! thanks for the PR @smiroddi √

saveriomiroddi:
Thanks to you, too! :+1:

PR #268, opened 2017-09-03
Recognize `-a` cmd-line arg

bgwines:
Hi, there! If there's a more appropriate place to make this change than the `Controller` ctor / if you'd like a test for this, please let me know :)

### DIFF
--- a/src/screenControl.py
+++ b/src/screenControl.py
@@ -296,11 +296,14 @@ class Controller(object):
             if not lineObj.isSimple():
                 self.lineMatches.append(lineObj)
 
-        self.numLines = len(lineObjs.keys())
-        self.numMatches = len(self.lineMatches)
-
         # begin tracking dirty state
         self.resetDirty()
+
+        if self.flags.args.all:
+            self.toggleSelectAll()
+
+        self.numLines = len(lineObjs.keys())
+        self.numMatches = len(self.lineMatches)
 
         self.setHover(self.hoverIndex, True)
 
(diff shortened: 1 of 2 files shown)


src/screenControl.py
```diff
@@ -296,12 +296,15 @@ def __init__(self, flags, keyBindings, stdscr, lineObjs, cursesAPI):
             if not lineObj.isSimple():
                 self.lineMatches.append(lineObj)
 
-        self.numLines = len(lineObjs.keys())
-        self.numMatches = len(self.lineMatches)
```
bgwines: had to move these because we should call `toggleSelectAll()` before assigning these, and need to call `resetDirty()` before calling `toggleSelectAll()`

pcottle:
Thanks for the PR @bgwines ! Do you have example workflow where this is helpful? Its the first I've had a request for `-a` so I'm just curious :P

This seems pretty clean though, think it's fine to merge even without a test. I'll try to add one though

pcottle:
Whoops, didn't make a PR for that but just pushed to master. But that should add a test for this :)

bgwines:
@pcottle, thanks so much! Yeah; my use-case is the following: when switching between branches I want to open all the files that were "relevant" last time I was on it, so I can more easily resume that work. An approximation for that set of files is the files that were modified in the last commit on that branch, so I would run `git show --name-only | fpp -c subl`, then cmd-A, then enter. With this commit I can pass in the `-a` flag (I tried running it with `-ni -ac subl`, which would be even better because it cuts out another keystroke, but it didn't seem to work for some reason; perhaps I'm misunderstanding `-ni`).

In an ideal world I could add some sort of hook to get Sublime Text to recognize that I switched branches and have it run this automatically; I should look into that :P

(I think Sublime Text has a concept of "projects" but I don't think it's perfectly suited to this workflow(?) based on preliminary fooling-around and this change is editor-agnostic anyway)

Have a nice weekend!

pcottle:
Ah cool, thanks for the extra info. And yeah we already have the “A” keystroke to select everything, but if you’re using FPP as part of a pipeline then eliminating keystrokes is always nice :P

Have a nice weekend too!

PR #275, opened 2017-12-27
enable split-by-default for nvim

julesferreira:
enable split-by-default for nvim


pcottle:
there we go! thanks for the update

PR #277, opened 2018-01-29
Use standard XDG directory structure; Fix #269

### LINKED ISSUE #269: Use $XDG_CACHE_HOME
Since #17 is fixed, all `fpp` files live in `~/.fpp`. It would be nice if `fpp` instead defaulted to the standard XDG directory structure used by many other projects:

https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

Since `~/.fpp` seems to store temporary cache data rather than configuration data, I'd like to propose moving it to `$XDG_CACHE_HOME/fpp/`. That is, `fpp` would use the first of the following locations that it finds:

1. `$FPP_DIR` (if `$FPP_DIR` is set).
2. `$XDG_CACHE_HOME/fpp/` (if `$XDG_CACHE_HOME` is set).
3. `~/.cache/fpp/` (default XDG cache dir, per spec)

(If you're comfortable removing `$FPP_DIR` in favor of the XDG convention, I think that it would be fine to eliminate #1. In that case, `fpp` can print a deprecation warning if it detects that `$FPP_DIR` is set.)

I understand that moving folders might not be a trivial choice, but I thought I'd at least ask for it if I care enough to work around it. :-)

ffledgling:
Change the default value of `$FPP_DIR` from `~/.fpp` to:

`$XDG_CACHE_HOME/fpp` if XDG_CACHE_HOME is set,
otherwise fall back to `~/.cache/fpp`.

This patch does *not* deprecate FPP_DIR.

Feedback requested: The patch does not add a warning about the change of cache path, I'm unsure if this has any impact on the operation (presumably not, since it's cache location), my educated guess is that it's non-critical and does not warrant a warning. If a warning is considered necessary I can add that in as well.

### DIFF
--- a/fpp
+++ b/fpp
@@ -23,8 +23,14 @@ BASEDIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 PYTHONCMD="python"
 NONINTERACTIVE=false
 
+# Setup according to XDG/Freedesktop standards as specified by
+# https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html
 if [ -z "$FPP_DIR" ]; then
-  FPP_DIR="$HOME/.fpp"
+  if [ -z "$XDG_CACHE_HOME" ]; then
+    FPP_DIR="$HOME/.cache/fpp"
+  else
+    FPP_DIR="$XDG_CACHE_HOME/fpp"
+  fi
 fi
 
 function doProgram {
--- a/src/stateFiles.py
+++ b/src/stateFiles.py
@@ -9,7 +9,7 @@ from __future__ import print_function
 
 import os
 
-FPP_DIR = os.environ.get('FPP_DIR') or '~/.fpp'
+FPP_DIR = os.environ.get('FPP_DIR') or '~/.cache/fpp'
 PICKLE_FILE = '.pickle'
 SELECTION_PICKLE = '.selection.pickle'
 OUTPUT_FILE = '.fpp.sh'
--- a/src/usageStrings.py
+++ b/src/usageStrings.py
@@ -133,7 +133,7 @@ files into panes for vim clients (aka sequential editing).
 
 PathPicker saves state files for use when starting up, including the
 previous input used and selection pickle. By default, these files are saved
-in ~/.fpp, but the $FPP_DIR environment variable can be used to tell
+in $XDG_CACHE_HOME/fpp, but the $FPP_DIR environment variable can be used to tell
 PathPicker to use another directory.
 
 ~ Colors ~


pcottle:
why the change to HOME/.cache? as I'm reading this diff, this wont be a no-op when the version gets upgraded right?

I just dont want people to lose their settings when they upgrade

ffledgling:
@pcottle The straightforward answer to your question is... because the spec says so. :)

Quoting verbatim from [the spec](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html)

```
$XDG_CACHE_HOME defines the base directory relative to which user specific non-essential data files should be stored. If $XDG_CACHE_HOME is either not set or empty, a default equal to $HOME/.cache should be used.
```
Based on the spec, and as pointed out in issue #269 , the correct default then, would be `$HOME/.cache/fpp`.

My understanding of the flow is that the `$FPP_DIR` is simply used to store the temporary pickle file generated by the python module and the lifecycle of the file is no more than the duration of running your command piped via `fpp` and making your selection. If I have this right, after the selection is made, the file is no longer useful and can be discarded.

I found no mention of settings in the README and didn't find any mention of permanent configs when looking through the code either. Do you have an example of something that'd be lost?

pcottle:
The selection files are used across multiple launches of `fpp`. If you run fpp with no input, it will use the last input and selection it was launched with.

so if someone has a super huge amount of input that they are using fpp with and upgrade in the middle of a delicate workflow, they might lose state/progress. it would be an edge case but I want to be respectful, especially if we can anyways :P

also isnt `-z` the null check? right now the code says "if there is no FPP_DIR, make it as `$HOME/.fpp`. but your code then changes that to "if there is no FPP_DIR and no XDG_CACHE_HOME, make it `$XDG_CACHE_HOME/fpp` which seems wrong?

ffledgling:
@pcottle Nice catch with the if condition, I fixed and re-pushed.

Re: the edge case, you're right about the case itself of course. I see three options here:

1. Add a warning on the first invocation of the new version somehow. This might not be trivial to do "correctly" however.
2. Silently migrate data from `~/.fpp` to the new location, somehow providing a seamless transition. Prior experience with automatic migrations tells me it's very likely we miss some kind of edge case though.
3. Decide that this use case is rare enough that we can safely ignore it, possibly with a mention in the release notes. If we want to be "correct" semver-wise, we release a new major version, although admittedly, that feels like overkill.

pcottle:
Alright I guess this can be good to go. Agreed its a bit of an edge case but it should only be disruptive on version upgrade, and work fine after that.

pcottle:
🎉  thanks for the fix and making this adhere to the spec!

ffledgling:
@pcottle Thanks!

PR #280, opened 2018-02-18
Fix typo: acceptible->acceptable

ismith:


pcottle:
Thanks for the fix! 🎉

PR #283, opened 2018-03-31
Match filename without directory but with number

Ngo-The-Trung:
A very common scenario that fpp doesn't match with a line number is when grep-like tools (grep, ripgrep) finds a file in the same directory it's run from and doesn't include a directory in the output:

```
$ grep -n my_pattern A.txt B.txt
A.txt:100 my_pattern
```

Before this commit this would match A.txt (`JUST_FILE` regex) which doesn't include the line number. This fixes this by adding `JUST_FILE_WITH_NUMBER`.

There are some other changes to make it pass autopep8

### DIFF
--- a/src/__tests__/testParsing.py
+++ b/src/__tests__/testParsing.py
@@ -61,7 +61,7 @@ fileTestCases = [{
 
     # For now, don't worry about matching the following case perfectly,
     # simply because it's complicated.
-    #}, {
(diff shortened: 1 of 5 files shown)


pcottle:
Wow thanks so much for adding functionality for your PR AND including test coverage! What a dream PR to accept :D 

🎉

PR #285, opened 2018-04-13
Added support for Vim and Emacs temporary/backup/save files

anshuman23:
Hello! 

So while this might be a trivial requirement, I feel that PathPicker should support the selection of Vim and Emacs backup files as well. These files are of the following two types: 
- Emacs save files (`example.txt~`)
- Vim and emacs backup files (`#example.txt#`)

 These files are usually automatically created by the respective text editors and serve as a sort of checkpoint to restore the original files they were created from (usually with the same name and without the tilde or pound). It makes sense to enable PathPicker to select such files, as users might be required to view them and/or edit them.

I have also written tests for these additions to `parse.py`. 

Cheers!

### DIFF
--- a/src/__tests__/testParsing.py
+++ b/src/__tests__/testParsing.py
@@ -155,6 +155,14 @@ fileTestCases = [{
     'match': True,
     'file': 'So.many.periods.txt'
 }, {
+    'input': 'So.many.periods.txt~',
+    'match': True,
+    'file': 'So.many.periods.txt~'
+}, {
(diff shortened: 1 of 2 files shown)


pcottle:
Awesome! 🎉  Typically I try to merge this logic into the other regexes but for something this specific / odd, a standalone regex makes sense (especially if it doesnt already match another one higher up in the tree)

thanks so much for contributing to path picker. I'm pretty tied up with some instagram work right now but I'll try to get a new release out sometime soon

PR #288, opened 2018-07-09
Add "execute keys" functionality

### LINKED ISSUE #287: Add functionality for executing keys when the list shows up
A functionality that would be helpful for a use case of mine (and that could be useful in general, for `fpp` scripting) is to specify keys to execute when launching fpp.

Specifically, I use fpp with the content of the terminal, and I'd like to immediately move to the bottom of the list, where the relevant content is.

This could be accomplished simply by adding something along the lines of `--execute-keys END`.

What do you think? I actually have a [working version](https://github.com/facebook/PathPicker/compare/master...saveriomiroddi:sav-add_execute_keys), but I'd like to 1. know if this can be accepted, and if so, 2. to get more input before writing the UTs.

saveriomiroddi:
(the functionality is actually implemented. I've marked this as WIP in order to start the feedback cycle; once it's completed, I'll add the UTs)

This functionality allow the user to specify a sequence of keys to execute when the files list shows up.

This is useful on certain cases, e.g. using "END" in order to automatically go to the last entry when there is a long list.

Closes #287.

### DIFF
--- a/src/screenControl.py
+++ b/src/screenControl.py
@@ -364,9 +364,9 @@ class Controller(object):
         self.moveCursor()
         while True:
             if len(executeKeys) > 0:
-              inKey = executeKeys.pop(0)
+                inKey = executeKeys.pop(0)
             else:
-              inKey = self.getKey()
+                inKey = self.getKey()
             self.checkResize()
             self.processInput(inKey)
             self.processDirty()
--- a/src/screenControl.py
+++ b/src/screenControl.py
@@ -356,12 +356,17 @@ class Controller(object):
         self.helperChrome.outputDescription(self.lineMatches[self.hoverIndex])
 
     def control(self):
+        executeKeys = self.flags.getExecuteKeys()
+
         # we start out by printing everything we need to
         self.printAll()
         self.resetDirty()
         self.moveCursor()
         while True:
-            inKey = self.getKey()
+            if len(executeKeys) > 0:
+              inKey = executeKeys.pop(0)
+            else:
+              inKey = self.getKey()
             self.checkResize()
             self.processInput(inKey)
             self.processDirty()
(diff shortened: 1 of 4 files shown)


pcottle:
This is a pretty minimal change! i think you just want to run pep formatting to fix the build:

```
         while True:
             if len(executeKeys) > 0:
-              inKey = executeKeys.pop(0)
+                inKey = executeKeys.pop(0)
             else:
-              inKey = self.getKey()
+                inKey = self.getKey()

```

also for the unit tests, we already have unit tests with mocked input. lets just get test coverage by sending it via this command flag (as close as we can get to mocking that) rather than the way its done now

saveriomiroddi:
Thanks :smile: :+1: 

I'll complete this with some delay, will complete PR+cleanup in around 10 days!

saveriomiroddi:
Ah, also, do you think execute keys is an appropriate description? An alternative I can think of is "playback keys", to better mirror the "record" functionality.

pcottle:
I think execute is fine, playback implies historical but you might be executing this fresh

pcottle:
Great!

saveriomiroddi:
@pcottle ready for review! if the review is OK, let me know if you want me to squash the commits before you merge (it won't be needed if your workflow is to squash merge).

src/screenControl.py
```diff
@@ -356,12 +356,17 @@ def describeFile(self):
...
     def control(self):
+        executeKeys = self.flags.getExecuteKeys()
+
         # we start out by printing everything we need to
         self.printAll()
         self.resetDirty()
         self.moveCursor()
         while True:
-            inKey = self.getKey()
+            if len(executeKeys) > 0:
```
pcottle: think

```
if executeKeys:
```

should suffice

PR #291, opened 2018-11-13
Fix line number when EDITOR contains arguments (e.g "emacs -nw")

guygadebois:
My EDITOR env variable is set to `emacsclient -nw`. Without this fix, all files open at line 1.
This fixes the detection of editors using the `+<line-number>` syntax, when they're invoked with other arguments.


pcottle:
looks good! good point that we probably wanted to split this all along :) 🎉

Issue #295, opened 2018-11-26
Command Mode not working in Python3


edVonSchleck:
I got an error when tyring to use command mode (Python 3.6):
Line 150 in pathpicker/src/output.py:
bytes object has no attribute encode

Could you fix that?

Thanks.

david-caro:
I seem to be unable to reproduce, tried with python 3.6.3rc1+ on CentOS 7.5 and with python 3.7.1 on Fedora 29.

I'm using the latest master a56e0530e6cbd6bdb9c0eed5789df3952dcb3daa, updated on Nov 21st.

Can you provide a bit more info about the environment you are having the issue in?

david-caro:
Thanks @RicardoMing, I'll look into it, but it looks like it's related to the version of python, can you try updating to a newer one? (>3.6)

david-caro:
Just tried installing anaconda, and setting up python 3.6.5, but was unable to reproduce:

Downloaded and installed the latest anaconda (from https://www.anaconda.com/download/#download), then ran the command:
```
bin/conda create -n 365 python=3.6.5 anaconda
```
Activated the venv:
```
source bin/activate 365
```
And made sure the python is redirected to the same version:
```
$ python --version
Python 3.6.5 :: Anaconda, Inc.
```

But got no error:
```
$ git status | ./fpp
No lines matched!!
```
tried also with something that generated some lines without problems:
```
$ git diff HEAD^^ | ./fpp
nothing to do!
```

@RicardoMing @edVonSchleck can you try running with the PR https://github.com/facebook/PathPicker/pull/296 ? 

That should give us more info. Thanks!

RicardoMing:
This is the new output with Anaconda Python3.6.5.
```
debug_295 $ ls | ./fpp
Traceback (most recent call last):
  File "/home/sjm/GitHub/PathPicker/src/output.py", line 151, in composeFileCommand
    command = command.encode().decode('utf-8')
AttributeError: 'bytes' object has no attribute 'encode'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/sjm/GitHub/PathPicker/src/choose.py", line 104, in <module>
    curses.wrapper(lambda x: doProgram(x, flags))
  File "/home/sjm/anaconda3/lib/python3.6/curses/__init__.py", line 94, in wrapper
    return func(stdscr, *args, **kwds)
  File "/home/sjm/GitHub/PathPicker/src/choose.py", line 104, in <lambda>
[...]

PR #296, opened 2019-01-03
DO NOT MERGE: added dummy exception raise for debugging #295

david-caro:


pcottle:
so we shouldnt merge this? will close it out then

PR #299, opened 2019-01-31
[output] Better handling of byte strings on execComposedCommand

### LINKED ISSUE #295: Command Mode not working in Python3

david-caro:
This fixes #295 by making sure any byte-string is decoded before
manipulating it.


pcottle:
seems straightforward, thanks for the install updates as well!

PR #306, opened 2019-08-30
deb file was not in most recent release assets

markstos:
Update README to reflect that building locally needs to be done on debian-based Linux systems


pcottle:
Thanks for the fix! Yeah I was never able to get the debian build locally myself, I think deferring to local build is the right call

markstos:
I just tested the local build option and it worked fine (On Debian 8).

PR #313, opened 2020-05-20
Change path of output file to match debian standard

ssgelm:
The standard debian filename of a package is PACKAGENAME_VERSION_ARCH.deb. Therefore, ideally it should be named pathpicker_0.9.2_all.deb. In the Debian manual at https://www.debian.org/doc/manuals/debian-faq/pkg-basics.en.html#pkgname, it says:

    The Debian binary package file names conform to the following
    convention: <foo>_<VersionNumber>-<DebianRevisionNumber>_<DebianArchitecture>.deb

This currently adds an extra step before we can add it to our local repo so it would be nice if this output the expected filename.

### DIFF
--- a/debian/package.sh
+++ b/debian/package.sh
@@ -47,7 +47,7 @@ find . -type f -exec chmod 644 {} \;
 echo "Building package..."
 rm "$PTH/package.sh"
 chmod 755 usr/share/pathpicker/fpp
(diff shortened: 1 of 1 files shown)


ssgelm:
Hilariously, I just realized that the poor filename was actually my fault in 2016 in #218 😬

pcottle:
No worries, thanks for following up and updating the name! Apologies for the delay in reviewing this PR :)

PR #314, opened 2020-06-06
add FB used repos

dragonxlwang:
at FB, those are commonly used repos;


pcottle:
Im going to merge this in addition to #315 so we dont have to set an environment variable by default to get sane behavior in FB land

PR #315, opened 2020-06-06
allow change REPOS by environment variables

dragonxlwang:
1 liner, and this is to allow customize repos; Either this or #314 can be used to unblock; Thanks


pcottle:
Good call! yeah I guess its easier to use an environment variable rather than edit the source code :P

PR #316, opened 2020-07-05
allow other editors to receive a file line number

deelam:
Current code only provides parsed file numbers to certain editors. This update will allow any editor to received the line number if environment variable FPP_LINENUM_SEP is set, e.g., to `:` or ` +`.


pcottle:
Oh wow sorry about missing this! My notifications got buried in another folder. Thanks for the fix and this looks great :)

PR #322, opened 2021-02-09
Fix bytes to string conversion

KapJI:
`subprocess.check_output` returns `bytes` and they should be converted to `str` with `decode()`.

These are all occurences where `subprocess.check_output` is used.

Before:
Screenshot 2021-02-09 at 01 03 10

After:
Screenshot 2021-02-09 at 01 02 48


pcottle:
🎉  thanks a bunch!

PR #397, opened 2021-02-24
Add support for the `micro` text editor

JonathanAquino:
This PR simply adds support for the `micro` text editor. This is a less well known editor that is growing in popularity. It takes the same input arguments as vi, emacs, etc.


KapJI:
Thanks!

PR #398, opened 2021-02-24
Don't keep the old selection if --keep-open is specified

JonathanAquino:
If `--keep-open` is specified, we should clear the old selection.

### Motivation

I like to use ack with fpp, like this:

    ack foo | fpp --keep-open

But since fpp doesn't clear the old selection, after I open a file (say, the 12th file) and close it, when fpp reopens, the 12th file is selected. So I need to go down to the 12th file and unselect it before I can choose a new file to open.

Instead, we should simply clear the old selection every time fpp reopens in `--keep-open` mode.

### DIFF
--- a/src/process_input.py
+++ b/src/process_input.py
@@ -77,6 +77,13 @@ def main(argv: List[str]) -> int:
         print(f"Done! Removed {len(state_files.get_all_state_files())} files ")
         return 0
     if sys.stdin.isatty():
+        # don't keep the old selection if the --keep-open option is used;
+        # otherwise you need to manually clear the old selection every
+        # time fpp is reopened.
+        if flags.get_keep_open():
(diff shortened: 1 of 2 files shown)


src/process_input.py
```diff
@@ -77,6 +77,14 @@ def main(argv: List[str]) -> int:
         print(f"Done! Removed {len(state_files.get_all_state_files())} files ")
         return 0
     if sys.stdin.isatty():
+        # don't keep the old selection if the --keep-open option is used;
+        # otherwise you need to manually clear the old selection every
+        # time fpp is reopened.
+        if flags.get_keep_open():
+            # delete the old selection
```
JonathanAquino, line 84: This is the same as the `delete the old selection` lines below.

KapJI:
It works, thanks!

Issue #403, opened 2021-05-30
Feature Request: Make man page easier to read, put each flag in its own line.


thiagowfx:
Original title: **Feature Request: Fpp persistent mode, for opening files in $EDITOR, one-by-one**.

This FR is a bit hard to describe, so I'll give a concrete example which is probably easier to understand.

Consider `pacman -Ql fpp` which prints the list of all files installed by the `fpp` package in Arch Linux:

```
% pacman -Ql fpp
fpp /usr/
fpp /usr/bin/
fpp /usr/bin/fpp
fpp /usr/share/
fpp /usr/share/doc/
fpp /usr/share/doc/fpp/
fpp /usr/share/doc/fpp/CODE_OF_CONDUCT.md
fpp /usr/share/doc/fpp/CONTRIBUTING.md
fpp /usr/share/doc/fpp/README.md
fpp /usr/share/doc/fpp/assets/
fpp /usr/share/doc/fpp/assets/ZeroClipboard.min.js
fpp /usr/share/doc/fpp/assets/ZeroClipboard.swf
fpp /usr/share/doc/fpp/assets/bootstrap.min.css
...
```

Now let's say I want to use `fpp` against this very output:

```
% pacman -Ql fpp | fpp
```

This does what you expect: It opens `fpp` against the output above.

Now here's my use case: I'd like to interactively navigate that output to open detected files one by one with my favorite `$EDITOR`:

1. I'd navigate to `/usr/bin/fpp` and press `Enter`. That opens it in my `$EDITOR`.
2. After studying the file, I quit the editor.

Now here's my **feature request**:

**Observed behavior**: After 2), I get back to the command prompt.

**Wanted behavior**: After 2), I would like to get back to the previous `fpp` exec, as if I had run `pacman -Ql fpp | fpp` again. My next action would be to use the `fpp` selection tool to open another file in my `$EDITOR` (say, `/usr/share/doc/fpp/README.md`), then close it, then open another file, and so on.

In other words, the idea is to maintain the output of a given command fixed, and then make `fpp` go continuously back to that output whenever the editor is closed.

In that scenario, it would be best if `fpp` didn't execute the command (`pacman -Ql fpp`) again, but rather just saved the output once and then kept going back to it.

This could be used for running commands multiple times as well; in my use case above, I open my `$EDITOR` multiple times, but I could also want to run different commands several times in a row, with the same output.

thiagowfx:
[...]

PR #404, opened 2021-07-01
Fix man page (#403)

thiagowfx:
- Updates man page with newly added flags since 2015. Man page had been
  last updated only in 2015.
- Fix formatting of flags in man page. Instead of joining lines, now
  each flag is correctly displayed in its own line, making it easier to
  read.
- Fix man page generation script, which was broken and still referencing
  the old usageStrings.py (correct filename is usage_strings.py).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ fpp.deb
 fpp*.deb
 debian/usr/
 pathpicker*.deb
+# man page leftovers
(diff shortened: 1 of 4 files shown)


KapJI:
Please format your code using black.

thiagowfx:
Done.

KapJI:
Thanks!

PR #416, opened 2022-02-19
remove homebrew formula

thiagowfx:
It is already maintained upstream:
https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/fpp.rb


pcottle:
I think I had the formula in here for easy copying when we need to update the recipe upstream

So removing the file might break this build:
https://github.com/facebook/PathPicker/blob/main/scripts/makeDist.sh#L11

since we need the template

thiagowfx:
Did you use this script when you created 0.9.5?

My take is that you don’t need to worry about it; packaging and upgrades are usually deferred to the community. Deleting the script would make your life as a maintainer easier by making it explicit the package isn’t maintained from this repo.

That said, this is just a suggestion. Keeping the script around certainly won’t hurt.

thiagowfx:
I’ll go ahead and close this then, but if you change your mind I could also remove the script you mentioned before submitting this PR.

PR #419, opened 2022-03-21
Ignore reserved characters in quick select mode LABELS

### LINKED ISSUE #410: Skip the `F` key in quick select mode
In quick select mode, I can never select the path in line `F`, because `F` is by default the "un-select" key.

cibinmathew:
Fixes #410

**Reserved**
A: toggle selection for all
F: toggle selection for current row


KapJI:
Thanks!

PR #421, opened 2022-06-08
Updating the README's debian dpkg file name

kristopolous:
The build process outputs `pathpicker` and not `fpp`. Very minor change but might as well be accurate.


kristopolous:
I signed the paperwork under my username here + the yahoo domain

KapJI:
Thanks!

PR #430, opened 2023-05-14
Synchronize pyproject.toml version with src/version.py

Jasha10:
The version number in `pyproject.toml` was out of sync with `src/version.py`.


thiagowfx:
LGTM (I forgot to bump it on https://github.com/facebook/PathPicker/pull/415)

PR #434, opened 2023-10-21
Fix failing CI builds

cibinmathew:
Recent invocations of CI on merge to master is failing as `ubuntu-latest` is missing Python3.6. Pinning to `20.04` to fix the builds


cibinmathew:
image

pcottle:
Thanks for the fix!

